PR #70, opened 2024-10-28
feat: add Catalan locale to the list of supported languages

sorydi3:
## Description

- added Catalan to the list of supported languages

Fixes # (issue number)

## Platform Affected

- [x] iOS
- [x] Android

## Checklist

  - [x] iOS Simulator/Device
  - [x] Android Emulator/Device


src/utils/l10n.ts
```diff
@@ -459,4 +459,63 @@ export const l10n = {
...
+    goToModels: 'Anar a Models',
+    readyToChat: "Preparat per xatejar? Carregueu l'últim model utilitzat.",
+    pleaseLoadModel: 'Si us plau, carregueu un model per començar a xatejar.',
+    shortMemoryWarning: 'Advertència de memòria',
+    memoryWarning:
+      'Advertència: La mida del model està prop o supera la memòria total del dispositiu ({{totalMemory}} GB). Això pot causar comportaments inesperats com baix rendiment o inestabilitat del sistema.',
+    load: 'Carregar',
+    cancel: 'Cancel·lar',
+    download: 'Descarregar',
+    offload: 'Descarregar',
```
a-ghorbani, line 482: I noticed both offload and download used the same term. I don't speak Catalan, so don't know if this is a mistake, or Descarregar would fit for both?

a-ghorbani (commented):
Thanks. I have only one comment.

src/utils/l10n.ts
sorydi3: It means the same in this context, just like in Spanish, as both words can be translated as 'descarregar'. However, we could use the word 'alliberar' for 'offload' to avoid confusion.

a-ghorbani: Cool. Thanks.

PR #142, opened 2024-12-18
[Feat] add benchmarking page.

### LINKED ISSUE #141: [Feat]: Model Benchmarking
**Description**

Benchmarking directly on a phone allows for a much better understanding of performance-related improvements.
Users can obtain more accurate information about what to expect from a model in terms of performance.
Additionally, we can enable users to share their benchmarks with the community on the soon-to-be-published "[AI Phone Leaderboard](https://huggingface.co/spaces/a-ghorbani/ai-phone-leaderboard)."

**Use Case**
The user might want to systematically compare the performance of two models on a phone. To achieve this, they would require a consistent method for processing prompts and generating text.

a-ghorbani:
## Description

This PR introduces a new benchmark page that enables users to compare the performance of models directly on their phones. It provides performance metrics and the ability to share results with the community.

**Benchmark Page**
- Prompt Processing (PP) tokens per second
- Text Generation (TG) tokens per second
- RAM peak usage

**Community Sharing**
Legit users, verified through app check, can share their benchmark results with the community. In the future we need to find a way to let users outside Play Store/Google Play be able to share too.

Results can be shared to the "[AI Phone Leaderboard](https://huggingface.co/spaces/a-ghorbani/ai-phone-leaderboard)."

Simulator Screenshot

Fixes #141 

## Platform Affected

- [x] iOS
- [x] Android

## Checklist

  - [x] iOS Simulator/Device
  - [x] Android Emulator/Device


a-ghorbani:
builds will fail if no firebase configs are available. This need to be fixed #143

Issue #166, opened 2025-01-10
[Bug]: Can't reliably delete chats


Laince1212:
The gesture to delete chats is very unreliable, sometimes it works sometimes it doesn't, can there be a reliable way to easily delete chats in the next update?

PR #185, opened 2025-01-27
feat: add session renaming functionality and context menu in sidebar

MSBaigO:
## Description
Added a menu on top of each of session where you can select whether to rename or delete the session.

Fixes: #166 

## Platform Affected

- [ X] iOS
- [ X] Android

image
image

### DIFF
--- a/src/components/SidebarContent/SidebarContent.tsx
+++ b/src/components/SidebarContent/SidebarContent.tsx
@@ -1,19 +1,23 @@
 import React, {useEffect, useState} from 'react';
-import {TouchableOpacity, View, TextInput, Modal, Button} from 'react-native';
-import {SafeAreaView} from 'react-native-safe-area-context';
+import {TouchableOpacity, View, TextInput, Modal} from 'react-native';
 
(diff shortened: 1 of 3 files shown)


a-ghorbani:
Thanks! Looks good. Just so you know, we are adding a menu in the chat page for deleting, etc. But I think having that functionality on the sidebar is nice too, so folks can delete without needing to navigate to the chat page.

Just one point, when renaming, it feels more natural if the text input is pre-filled with the old chat session title.

MSBaigO:
> Thanks! Looks good. Just so you know, we are adding a menu in the chat page for deleting, etc. But I think having that functionality on the sidebar is nice too, so folks can delete without needing to navigate to the chat page.
> 
> Just one point, when renaming, it feels more natural if the text input is pre-filled with the old chat session title.
I updated to include current title in rename window. Using the sidebar for renaming and deleting chat sessions seems to be the norm for most of the AI chats I have seen.

a-ghorbani:
amazing! This can be merged. Just fixed a few linting issues.

PR #201, opened 2025-02-04
Feat/chat UI improvements

Mukhammadali:
## Description

Changes:
- Updated Menu component styles
- Added new Chat Menu with chat generation settings, model selection, rename, delete, duplicate functionalities
- Updated Model settings UI
- Updated Chat Header which now displays active model

## Platform Affected

- [x] iOS
- [x] Android

## Checklist

  - [x] iOS Simulator/Device
  - [x] Android Emulator/Device


src/components/Menu/MenuItem/MenuItem.tsx
a-ghorbani, line 1: It looks like the submenu width is restricted by its parent menu? Is there an easy way to allow the submenu to be wider than the parent? The model names need more space, and we have extra space available to display them. It is a minor issue, it we don't have an easy fix, let's skip for now.

Screenshot 2025-02-05 at 07 55 36

a-ghorbani: Alternatively, if we use a different visual cue (highlight, bold, etc) to show which model is selected, we can remove the leading space reserved for the icon in the model list. This would free up enough space as well.

Mukhammadali: done
image

src/components/ChatGenerationSettingsSheet/index.ts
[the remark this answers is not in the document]
Mukhammadali: done

PR #207, opened 2025-02-14
fix: refactor downloader

### LINKED ISSUE #199: [General]: Where the downloaded GGUF placed?
Since I cannot directly download the gguf file from App's Huggingface button as the file always broken even I set the App to run in the background.  I then download the gguf file to my desktop and wanted to transfer it to Android -- data -- com.pocketpalai -- files.  However, the App cannot see the said file from the screen of the Android phone.

Version : 1.6.7

Any idea?

a-ghorbani:
## Description

Switching the download method from RNFS to [react-native-background-downloader](https://github.com/kesha-antonov/react-native-background-downloader), as it appears to be more stable, particularly for Android and background downloads.

Fixes #199 and #179 

## Platform Affected

- [ ] iOS
- [ ] Android

## Checklist

  - [ ] iOS Simulator/Device
  - [ ] Android Emulator/Device
- [ ] Unit tests and integration tests pass locally.

### DIFF
--- a/__mocks__/external/@kesha-antonov/react-native-background-downloader.js
+++ b/__mocks__/external/@kesha-antonov/react-native-background-downloader.js
@@ -1,0 +1,61 @@
+import {NativeModules} from 'react-native';
+
+// states:
+// 0 - Running
+// 1 - Suspended / Paused
+// 2 - Cancelled / Failed
+// 3 - Completed (not necessarily successfully)
+
+NativeModules.RNBackgroundDownloader = {
+  addListener: jest.fn(),
+  removeListeners: jest.fn(),
+  download: jest.fn(),
+  pauseTask: jest.fn(),
+  resumeTask: jest.fn(),
+  stopTask: jest.fn(),
+  TaskRunning: 0,
+  TaskSuspended: 1,
+  TaskCanceling: 2,
+  TaskCompleted: 3,
+  checkForExistingDownloads: jest.fn().mockImplementation(() => {
+    const foundDownloads = [
+      {
+        id: 'taskRunning',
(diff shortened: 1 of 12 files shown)


a-ghorbani:
we need this feat though before I can merge: https://github.com/kesha-antonov/react-native-background-downloader/pull/43

a-ghorbani:
### Android - Download Manager Testing
We need to verify that downloads work correctly in different app states and handle cancellations properly.

#### Foreground & Background Behavior
- [ ] Download starts and completes successfully when the app is in the foreground.
    - [ ] UI elements such as progress bar and ETA/speed are displayed correctly.
- [ ] Download continues and completes successfully when the app is in the background (not killed).
    - [ ] Verify that the download does not stop when moving the app to the background.  
    - [ ] When returning to the foreground, the download should still be progressing or completed successfully.  

#### Handling Cancellations
- [ ] Cancelling a download properly stops the download job.
    - [ ] The partially downloaded file should be removed.  
    - [ ] The download should not continue in the background as an orphan process.  

#### Handling App Termination (Killed State)  
- [ ] If the app is killed during an active download, handle it gracefully.
    - [ ] Ideally, (if we get the time) the download resume correctly.
    - [ ] otherwise, it should clean up any corrupted/incomplete files.

---

### iOS - Download Manager Testing
Since we have not changed the iOS backend module (RNFS), the main focus is verifying downloads work as expected.

#### Foreground & Background Behavior
- [ ] Download starts and completes successfully when the app is in the foreground.
- [ ] Download continues and completes successfully when the app is in the background (not killed).

#### Handling Cancellations
- [ ] Cancelling a download should stop the process and remove the partially downloaded file.

PR #307, opened 2025-05-12
feat: add backup and restore support to android

ginchly:
## Description

Add backup and restore support to Android, adding separate config for Android 12+, and older versions of Android

I've kept the exclusions consistent, but for Android 11 and below there is no way to specify exclusions only and that it must be encrypted, you can only do that on specific inclusions. It's not obvious to me what sufficient inclusions would be and that might get out of date over time, although perhaps just including some of the key directories could be sufficient?

Testing was done in emulators (Android 11 and Android 15) using steps from https://developer.android.com/identity/data/testingbackup. I used the bmgr commands and then did manual uninstall and install of the app, which triggers a restore event.

Fixes https://github.com/a-ghorbani/pocketpal-ai/discussions/293

## Platform Affected

- [ ] iOS
- [X] Android

## Checklist

- [X] Necessary comments have been made.
- [X] I have tested this change on:
  - [ ] iOS Simulator/Device
  - [X] Android Emulator/Device
- [X] Unit tests and integration tests pass locally.

## Testing
For Android 11:
* The standard commands using LocalTransport do not work when clientSideEncryptionFlag is added
* I tested by
  * Creating Pixel 4 / Android 11 emulator (note it must be an emulator with "Play" support)
  * Signing in with a google account and enabling device backup (offered after you sign in)
  * Setting a lock screen on the emulator (required for client side encryption)
  * `adb shell bmgr backupnow com.pocketpalai`
  * Uninstall and re-install the app
  * Check chat history, pals, settings restored (I checked dark theme restored correctly)

For Android 12 and above
* Android Studio preview has in-built backup and restore testing tools https://developer.android.com/studio/preview/features#test-backup-restore


android/app/src/main/res/xml/backup_rules_legacy.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<full-backup-content>
```
a-ghorbani, line 2: these are the files/folder we need.
```
    <include domain="sharedpref" path="." requireFlags="clientSideEncryption"/>
    <include domain="database" path="." requireFlags="clientSideEncryption"/>
    <include domain="file" path="profileInstalled" requireFlags="clientSideEncryption"/>
    <include domain="file" path="db-migration-complete.flag" requireFlags="clientSideEncryption"/>
    <include domain="root" path="pocketpalai.db" requireFlags="clientSideEncryption"/>
    <include domain="root" path="pocketpalai.db-wal" requireFlags="clientSideEncryption"/>
    <include domain="root" path="pocketpalai.db-shm" requireFlags="clientSideEncryption"/>
```

a-ghorbani:
Amazing. I took a look.
These are the files and folders that need to be included:
```
    <include domain="sharedpref" path="." requireFlags="clientSideEncryption"/>
    <include domain="database" path="." requireFlags="clientSideEncryption"/>
    <include domain="file" path="profileInstalled" requireFlags="clientSideEncryption"/>
    <include domain="file" path="db-migration-complete.flag" requireFlags="clientSideEncryption"/>
    <include domain="root" path="pocketpalai.db" requireFlags="clientSideEncryption"/>
    <include domain="root" path="pocketpalai.db-wal" requireFlags="clientSideEncryption"/>
    <include domain="root" path="pocketpalai.db-shm" requireFlags="clientSideEncryption"/>
```
I noticed I didn't put `pocketpalai.db` in the databases folder. We’ll need to move it there in the future, but for now we can leave it as is.

Do you mind to include these in your pr?

android/app/src/main/res/xml/backup_rules_legacy.xml
ginchly: thanks! updated

ginchly:
Thanks for your patience here - it took me a while to work out how to test this on Android 11, as the standard bmgr tools didn't work. I updated the description with the steps I took, seems to work as expected - no lock screen meant no backup, and when I added one it backed up successfully as that is the key for client side encryption. Chats, pals, settings seemed to restore correctly, and models weren't restored as expected.

PR #373, opened 2025-07-14
[Feat]: improve chinese localization translation (0715)

xiaoxing2009:
## Description

Refine the Chinese translation to align with linguistic preferences in China.
(again)

## Platform Affected

- [x] iOS
- [x] Android

## Checklist

  - [ ] iOS Simulator/Device
  - [x] Android Emulator/Device


a-ghorbani:
Thanks. I see the Japanese part has been changed too. Is that intentional?

xiaoxing2009:
> 谢谢。我看到日文部分也改了，是故意的吗？

**While translating the Chinese text, I also made adjustments to the Japanese version, for which I apologize.**
I originally assumed Japanese used the comma "，" for clause breaks, so I made minor adjustments to match my understanding. However, I later realized Japanese actually uses the symbol "、" for this purpose. I sincerely apologize for this mistake and will correct it immediately to align with standard Japanese conventions.
 
**If you prefer that I only focus on Chinese translations, I will remove all non-Chinese edits right away!**

LinXing0926:
> Thanks. I see the Japanese part has been changed too. Is that intentional?

**While translating the Chinese text, I also made adjustments to the Japanese version, for which I apologize.**
I originally assumed Japanese used the comma "，" for clause breaks, so I made minor adjustments to match my understanding. However, I later realized Japanese actually uses the symbol "、" for this purpose. I sincerely apologize for this mistake and will correct it immediately to align with standard Japanese conventions.
 
**If you prefer that I only focus on Chinese translations, I will remove all non-Chinese edits right away!**

a-ghorbani:
Thanks!

PR #419, opened 2025-08-29
Feat: Adding Russian localization support

Kuzmich55:
**Add Russian localization support**
This PR adds initial Russian language support to the application.

**Description of changes**
Added Russian translations for main interface strings
This is a draft version of the translation that requires further development

**Current status**
The translation is in development stage and needs:
- Terminology corrections
- Debugging and testing
- Overall localization quality improvements

**Important note**
Initially, the work was conducted through Weblate, however it was discovered that the source strings for translation in Weblate are outdated and do not correspond to the current state of the codebase. Therefore, it was decided to prepare this PR to create a baseline version of Russian localization.

**Next steps**
It is recommended to create a preview version of the application for community testing of the localization and gathering feedback.

Note: This is a foundational implementation. Community feedback and further iterations are welcomed to improve translation quality.


remixer-dec:
Can confirm that this translation is mostly correct with a few mistakes of machine translation.
Here are a few notes:
roleplayPalSheet.worldPlaceholder should be "Фэнтези", not "Фантазия"
generatingPrompt.label has been [renamed](https://github.com/a-ghorbani/pocketpal-ai/blame/1548e39fa818449399473c28745e968114ef13e2/src/utils/l10n.ts#L854) to "Purpose & Context" so it should be "Цель и контекст"
with the placeholder "Опишите роль, цель или сценарий для этого Pal...',

I understand that the creator of this app wants to use "Pal" branding, however in current translation the wording
"Pal" in all the strings is not using correct grammatical cases and is always refferred in nominative case. This approach is sometimes used for foreign words but it sounds very unnatural at least when we refer to Pal as a character and not as a template. In my opinion it would be better to add a case-ending latter after the word Pal, so it looks like this: 

exportAllPals: 'Экспорт всех Pal' -> 'Экспорт всех Pal’ов'
importPals: 'Импорт Pal' -> 'Импорт Pal’a'
palName: 'Имя Pal' -> 'Имя Pal’a'
disablePal: 'Выключить активный Pal'-> disablePal: 'Выключить активного Pal’a',

If Pal is referred more as a template than a character, current approach is fine. 
I'd also keep benchmarkResultCard.modelSettings.batch and benchmarkResultCard.modelSettings.ubatch with original text

a-ghorbani:
Thanks! appreciate this PR and the translations. I'm a bit slow on this. I'm switching over to proper i18n with Weblate, so there might be a bit of a delay here.

a-ghorbani:
Hi @Kuzmich55, thanks for the Russian translation work on both the PR and Weblate!

I've completed the Weblate migration and migrated the existing Russian translations to the new format. It's currently at ~48% coverage since new strings were added.

You can continue working directly on Weblate:
https://hosted.weblate.org/projects/pocketpal-ai/translations/ru/

Once it reaches 100%, we'll add Russian to the app.

PR #458, opened 2025-10-19
feat: add OCI registry support for pulling models from Docker Hub

ericcurtin:
docs: add comprehensive OCI models documentation


ericcurtin:
@a-ghorbani you might find this easier to test than me. I think this improves UX also, someone can select:

smollm2

and pull it from Docker Hub.

a-ghorbani:
@ericcurtin Thanks for this, I'm testing it now. I see the model list is quite long. I think we should keep the preset models shorter. I was even thinking about reducing the number of previously preset models.

Is there an API to search the models? That way, we could treat them like HF models to explore/search through. Or do you thin could we set up one?

PR #621, opened 2026-03-05
feat: add pinned/starred chat sessions

### LINKED ISSUE #602: [Feat]: Pinned/starred chat sessions
## Summary

Allow users to pin/star chat sessions so they appear at the top of the sidebar, separate from the date-grouped session list. Useful for ongoing projects or frequently-referenced chats.

## Motivation

The session list grows quickly and important chats get buried under the date grouping. A simple pin mechanism solves this without requiring full folder/tag support.

## Proposed Implementation

- Add `pinned: boolean` column to the `sessions` WatermelonDB table (schema migration)
- Long-press menu on a session gains a "Pin" / "Unpin" item
- Sidebar shows a **Pinned** section above the date-grouped sections when at least one session is pinned
- Pin state persists across app restarts

## Acceptance Criteria

- [ ] Pinned sessions appear in a dedicated section at the top of the sidebar
- [ ] Long-press menu allows pin/unpin toggle
- [ ] Pin state is persisted in the database
- [ ] Unpinned sessions continue to appear in the normal date-grouped list

sakebomb:
## Summary
- Add ability to pin chat sessions so they always appear at the top of the sidebar under a dedicated "Pinned" section
- Long-press a session in the sidebar to access the new Pin/Unpin option in the context menu
- Pinned sessions are persisted via a new `pinned` boolean column on `chat_sessions` (schema v7→v8)

## Details
- **Schema**: Adds `pinned` boolean column to `chat_sessions` table, bumps schema to v8 with migration
- **Store**: `groupedSessions` getter separates pinned sessions into a "Pinned" group rendered before date groups; `togglePinSession()` derives the target and persists via the repository's absolute `setSessionPinned(sessionId, pinned)` setter
- **UI**: Pin/Unpin menu item added to the session long-press context menu; pinned rows also carry a star so the state stays visible when section headers scroll out of view
- **l10n**: Adds `pin`, `unpin`, and `dateGroups.pinned` strings to `en.json`
- **Tests**: repository coverage for the new setter, store coverage for a pin surviving a reload, a component test for the menu item, a schema/migrations invariant test, and a locale guard keeping the sidebar group labels distinct

## Test plan
- [ ] Fresh install: verify migration creates `pinned` column, all sessions default to unpinned
- [ ] Long-press a session → verify Pin option appears in context menu
- [ ] Pin a session → verify it moves to "Pinned" section at top of sidebar
- [ ] Pin multiple sessions → verify they sort by date (newest first) within Pinned section
- [ ] Unpin a session → verify it returns to its date group
- [ ] Verify pinned state persists across app restarts
- [ ] Run `yarn test` — all tests pass
- [ ] Run `yarn lint` and `yarn typecheck` — no errors

Closes #602

### DIFF
--- a/__mocks__/repositories/ChatSessionRepository.js
+++ b/__mocks__/repositories/ChatSessionRepository.js
@@ -113,8 +113,9 @@ class ChatSessionRepository {
     return; // Mock: do nothing
   }
 
-  async toggleSessionPinned(sessionId) {
-    return true; // Mock: return new pinned state
+  async setSessionPinned(sessionId, pinned) {
+    this._pinned = this._pinned || {};
+    this._pinned[sessionId] = pinned;
   }
 
   // Update session completion settings
--- a/__mocks__/repositories/ChatSessionRepository.js
+++ b/__mocks__/repositories/ChatSessionRepository.js
@@ -113,6 +113,10 @@ class ChatSessionRepository {
     return; // Mock: do nothing
   }
 
+  async toggleSessionPinned(sessionId) {
+    return true; // Mock: return new pinned state
+  }
+
   // Update session completion settings
   async updateSessionCompletionSettings(sessionId, settings) {
     return; // Mock: do nothing
--- a/__mocks__/stores/chatSessionStore.ts
+++ b/__mocks__/stores/chatSessionStore.ts
@@ -89,7 +89,9 @@ export const mockChatSessionStore = {
   deselectAllSessions: jest.fn(),
   bulkDeleteSessions: jest.fn().mockResolvedValue(undefined),
   bulkExportSessions: jest.fn().mockResolvedValue(undefined),
+  togglePinSession: jest.fn().mockResolvedValue(undefined),
   dateGroupNames: {
+    pinned: 'Pinned',
     today: 'Today',
     yesterday: 'Yesterday',
     thisWeek: 'This week',
--- a/src/components/SidebarContent/SidebarContent.tsx
+++ b/src/components/SidebarContent/SidebarContent.tsx
@@ -18,6 +18,7 @@ import {
   PalIcon,
   SettingsIcon,
   ShareIcon,
+  StarIcon,
   TrashIcon,
   AppInfoIcon,
 } from '../../assets/icons';
@@ -33,11 +34,13 @@ const isDebugMode = __DEV__;
 interface SessionItemProps {
   session: SessionMetaData;
   isActive: boolean;
+  isPinned: boolean;
   onPress: (sessionId: string) => void;
   onLongPress: (sessionId: string, event: any) => void;
   menuVisible: string | null;
   menuPosition: {x: number; y: number};
   onMenuDismiss: () => void;
+  onPressPin: (sessionId: string) => void;
   onPressRename: (session: SessionMetaData) => void;
   onPressDelete: (sessionId: string) => void;
   onPressExport: (sessionId: string) => void;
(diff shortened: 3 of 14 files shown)


a-ghorbani:
## Visual evidence — pinned sessions (before/after the icon fix)






a-ghorbani:
Thanks for this — the feature is solid in the places that are easiest to get wrong: the migration is upgrade-safe, the layering runs one direction with no new cross-store coupling, and the grouping respects its ordering contract without mutating the session list.

I reviewed it in depth and pushed the fixes as one commit (`ebffdcc7`) rather than sending you a round-trip. Here's what changed and why.

### The pin icon painted nothing when a session was unpinned

`star.svg` is stroke-only, and `.svgrrc` rewrites its hardcoded `#333333` to the **`fill`** prop — so `fill="none"` zeroed the only painted attribute and the menu item showed a blank icon slot in the default state of every session. The pinned branch worked by accident (stroke *and* inherited fill both resolved to the primary colour). Fixed by passing the colour through the prop the asset actually honours and omitting it to get the outline. Also sized the glyph to 20 to match its siblings, fenced it off with a divider, and gave it a `testID`. Before/after captures are in the comment above.

Worth knowing for next time: `.svg` is mapped to a string in the jest config, so **no unit test can catch an icon that renders blank** — only a simulator capture will.

### `toggleSessionPinned` → `setSessionPinned(sessionId, pinned)`

Two defects fell out of the toggle shape. Both were reproduced with failing tests before the change:

- It returned `false` both for "now unpinned" and "record not found", and the store wrote that value straight onto the session. A failed lookup therefore unpinned a session the database still had pinned, with no error — and the pin came back on the next load.
- It read the current value *outside* `database.write`, so two overlapping toggles computed the same target and one tap was lost.

The absolute setter matches the three sibling mutators (`updateSessionTitle`, `setSessionActivePal`, `setSessionSettingsSource`) and drops the `as unknown as ChatSession` cast. The store now derives the target from its own state and mirrors it only after the write resolves.

### Pinned state on the row itself

`isPinned` was only consumed inside the menu, and the section headers aren't sticky — so once the list scrolled, nothing indicated which chats were pinned; you had to long-press and read whether the menu said Pin or Unpin. Pinned rows now carry a small star.

### Two judgment calls, so you know the reasoning

- **Pin failures stay silent.** Unlike export and delete, there's no alert. Pinning is non-destructive and trivially retryable, and the row not moving is itself honest feedback — a modal for "couldn't pin" costs more than it tells you. The part that did matter was the UI asserting an *unconfirmed* state, which the setter change fixes.
- **`pinned` is not exported.** The export format stays conversation-content-only. A pin is local organizational state, and importing someone else's chat shouldn't inherit their pins. (`settingsSource` is excluded on the same reasoning.)

### Tests added

Repository-level coverage for the new setter (the suite mocks the repository globally, so these go through `requireActual`), store coverage for a pin surviving a reload, a component test for the menu item, a schema/migrations invariant test, and a locale guard that keeps the sidebar group labels distinct.

### Verified

- `yarn lint` and `yarn typecheck` clean; full suite green at 265 suites / 3986 tests.
- The v7→v8 migration exercised on a simulator against a genuine pre-upgrade database: version bumped, all pre-existing rows preserved, and `pinned` backfilled to `0` with zero NULLs.

### One housekeeping note

The PR description still says schema v5→v6 and "93/93 passing"; the branch actually lands v7→v8. Worth updating so the description matches what merges.

Deferred and explicitly not blocking: the group map is keyed by the translated label, so a future translation collision could drop the Pinned group entirely (guarded by a unit test for now — the real fix is stable keys resolved at render time), and pinned rows would benefit from an accessibility label, which needs a new string.

a-ghorbani (approved):
lgtm

PR #624, opened 2026-03-07
feat: quick generation settings panel in chat

### LINKED ISSUE #605: [Feat]: Quick generation settings panel in chat
## Summary

Expose the most-used generation parameters (temperature, top-p, max tokens) in a compact panel accessible directly from the chat input row, without opening the full settings sheet.

## Motivation

Adjusting temperature or max tokens currently requires tapping the three-dot menu → Generation Settings → navigating a full sheet. For users who frequently tune these while chatting, this is too many steps.

## Proposed Implementation

- Small pill/chip in the chat input control bar showing the current temperature (e.g. `T: 0.7`)
- Tapping opens a compact bottom sheet with sliders for temperature, top-p, and max new tokens
- Changes apply to the current session's completion settings immediately
- Falls back to global new-chat settings when no session is active

## Acceptance Criteria

- [ ] Temperature visible in the input bar as a quick indicator
- [ ] Tapping opens a compact panel with temperature, top-p, max tokens sliders
- [ ] Changes persist to the session's completion settings
- [ ] No new native dependencies

sakebomb:
## Summary
- Adds a temperature pill (T: X.X) in the chat input control bar showing the current session temperature (shown when a session is active and a model is loaded).
- Tapping the pill opens a compact bottom sheet with sliders for temperature and top-p, and an **Unlimited toggle + max-tokens slider**.
- Changes persist to the session's completion settings immediately via `updateSessionCompletionSettings` (which flips `settingsSource` to `custom`).
- Reset to defaults restores the system defaults.
- No new native dependencies.

## Max tokens / unlimited
`n_predict` defaults to `-1` (unlimited) since #687 (with a migration flipping existing `1024 → -1`). A 64–8192 slider can't represent unlimited, so max tokens is an **Unlimited toggle** plus a finite slider — mirroring the full Generation Settings sheet's semantics. Turning the toggle off starts the slider at the same `1024` fallback the full sheet uses. This avoids the panel showing a false "64" cap on a default (unlimited) session and silently capping generation on any slider touch.

## Test plan
- [x] Temperature pill appears when a session is active and a model is loaded; hidden otherwise.
- [x] Tapping the pill opens the QuickGenSettingsSheet.
- [x] Sliders load the current session's completion settings.
- [x] Apply saves settings and closes the sheet.
- [x] Unlimited session → toggle on, slider hidden; toggle off → slider appears at 1024; apply preserves `-1` when left on, caps when turned off.
- [x] Reset restores the unlimited default (temp 0.7, top-p 0.95, max tokens unlimited).
- [x] No active session → falls back to defaults.
- [x] `tsc`, `eslint`, `l10n:validate` clean; full suite 269 suites / 4165 passed.
- [ ] Manual device pass (iOS/Android).

Closes #605

### DIFF
--- a/__mocks__/stores/chatSessionStore.ts
+++ b/__mocks__/stores/chatSessionStore.ts
@@ -6,10 +6,11 @@ import type {
 
 // Mock defaultCompletionSettings to avoid circular imports
 // This should match the actual defaultCompletionSettings from ChatSessionStore
+// (n_predict default is -1 = unlimited since #687).
 export const mockDefaultCompletionSettings = {
   version: 3,
   include_thinking_in_context: true,
-  n_predict: 1024,
+  n_predict: -1,
   temperature: 0.7,
   top_k: 40,
   top_p: 0.95,
--- a/src/components/ChatInput/ChatInput.tsx
+++ b/src/components/ChatInput/ChatInput.tsx
@@ -80,6 +80,8 @@ export interface ChatInputTopLevelProps {
   reasoningEffort?: string;
   /** Callback to cycle the graded effort state (off -> values -> off) */
   onEffortCycle?: () => void;
+  /** When provided, shows a temperature pill that opens the generation settings panel */
+  onGenSettingsPress?: () => void;
 }
 
 export interface ChatInputAdditionalProps {
@@ -105,6 +107,8 @@ export interface ChatInputAdditionalProps {
   reasoningEffort?: string;
   /** Callback to cycle the graded effort state (off -> values -> off) */
   onEffortCycle?: () => void;
+  /** When provided, shows a temperature pill that opens the generation settings panel */
+  onGenSettingsPress?: () => void;
 }
 
 export type ChatInputProps = ChatInputTopLevelProps & ChatInputAdditionalProps;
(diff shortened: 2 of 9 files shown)


sakebomb:
Rebased onto current `main` and did a self-review pass. One real defect surfaced that only appears against current `main`:

**`n_predict` now defaults to `-1` (unlimited)** (via #687 + a migration flipping existing `1024 → -1`). The original max-tokens slider was `64–8192`, so on a default session it displayed a false "64" cap, and any drag would silently cap an otherwise-unlimited session. Fixed by making max tokens an **Unlimited toggle + finite slider**, mirroring the full Generation Settings sheet (finite fallback `1024`).

Also corrected here: the shared test mock's `defaultCompletionSettings.n_predict` was stale at `1024` (now `-1`, matching the real default), and the sheet's tests were rewritten to actually exercise the unlimited/finite/toggle-off/reset paths (the old ones asserted a `1024` default that no longer exists). Both the toggle-save and slider-visibility guards are mutation-verified.

Verification: `tsc` clean · `eslint` clean · `l10n:validate` valid · full suite 269 suites / 4165 passed. No native changes.

PR #631, opened 2026-03-14
exposed controls, VLM import, prompt customization and related experiments

CCSSNE:
## Description

Since the majority of the users of this application are those who wish to gain a deeper understanding of large models, I place greater emphasis on learning and exploration. I try to provide more parameters for users to debug, with easy recovery from incorrect adjustments, while exposing as much of the pipeline as possible so users can clearly see the model’s inputs, outputs, and adjustment methods. Because the chain is fully controllable, when the model encounters issues, configurations can be adjusted directly instead of waiting for code-level changes, making experimentation more flexible.

This branch represents a working implementation that I have been using and testing on my side.

I intentionally squashed this into a single commit, since the original commit history mostly reflected the development process and was quite messy, making it unsuitable for review.

If some of the directions here seem useful or aligned with the project, I can help isolate and clean those parts into proper commits or smaller PRs later.

This fork mainly reflects my own usage needs.

Main changes in this fork include:

- Added the ability to directly import local VLM models
- Support for both model imports and visual component imports for MM Projects
- Fully exposed and customizable prompt template logic
- Added functionality to share GGUF model files
- Implemented logging features
- Exposed several interfaces that previously existed internally but were not visible in the UI
- Optimized the template matching mechanism in v132. Besides the 10 recommended built-in models, it now prioritizes reading the model's own configuration to reduce mismatching across different model families

One thing I originally intended to add but did not complete:

- real-time monitoring of RAM / virtual memory usage, CPU utilization, and power consumption for performance debugging. I dropped this for now because the implementation complexity became too high.

The README in my fork also summarizes the general direction of these changes.

<p align="center">
  
  
  
</p>

## Platform Affected

- [ ok] Android

## Checklist

  - [ ] iOS Simulator/Device
  - [ok ] Android Emulator/Device
- [ ok] Unit tests and integration tests pass locally.


CCSSNE:
Add a progress bar featuring location markers for each conversation, along with up/down navigation toggles to quickly jump between different conversations.

Real-time memory usage display has also been enabled on Android.

Added compilation for Apple IPA installers.

Split the single "Think" button into three buttons:
Whether to enable the UI-layer thinking box.
Whether to activate thinking at the parameter level.
Whether to incorporate thinking into the context.
Added an independent control for the visual encoder to freely choose between CPU or GPU loading, with CPU now set as the default. This improves compatibility compared to the original forced GPU mode, which caused various issues including crashes and image recognition errors.

PR #656, opened 2026-03-29
feat: remove the limit of N predict (#291)

### LINKED ISSUE #291: [Feat]: lift the limit of N predict
**Description**
Could you please lift the limit of N predict? it is now set at 4096 tokens. When I’m using models like Qwen3, it is a little bit too short.

**Use Case**
For example, when I’m solving a complex math question or discussing about current events and writing articles, 4000 tokens will be too short.

kumikumi:
## Description

Fixes #291 

Allow any value larger than 0 to be used as the N predict limit. This is especially useful with reasoning and other long running models.

## Platform Affected

- [X] iOS
- [X] Android

## Checklist

- [X] Necessary comments have been made.
- [X] I have tested this change on:
  - [X] iOS Simulator/Device
  - [ ] Android Emulator/Device
- [X/?] Unit tests and integration tests pass locally.

### DIFF
--- a/src/utils/__tests__/modelSettings.test.ts
+++ b/src/utils/__tests__/modelSettings.test.ts
@@ -85,6 +85,34 @@ describe('modelSettings', () => {
         'Please enter a valid number',
       );
     });
+
+    it('supports rules with only a minimum bound', () => {
+      const minOnlyRule = {
+        type: 'numeric' as const,
+        min: 1,
+        required: true,
+      };
+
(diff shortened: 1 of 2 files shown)


a-ghorbani:
how about making min=-1, and perhaps changing the default to -1 as well (which means unlimited)?

a-ghorbani:
Thanks for the contribution! #687 supersedes this one — it takes a similar approach but defaults to `-1` (unlimited, matching llama.cpp semantics) instead of just removing the cap.

PR #669, opened 2026-04-06
chore(ci): add llama.rn build-from-source support

a-ghorbani:
## Summary

- Auto-detect git ref vs npm version for `llama.rn` in CI and set build-from-source flags accordingly
- Add iOS `RNLLAMA_BUILD_FROM_SOURCE=1` env propagation and Android `rnllamaBuildFromSource` gradle property toggle in CI
- Cache Android `.cxx` NDK build artifacts for faster source builds
- Create `tools/use-llama-rn-source.sh` developer toggle script (supports commit SHA, fork refs, and npm versions)
- Update `postinstall.sh` to build JS output via `react-native-builder-bob` when `lib/` is missing (git installs)
- Add Podfile `post_install` hook for header search paths when building from source
- Point `llama.rn` to crash-fix commit `0ee4fe6` for testing

## Changed Files

| File | Change |
|------|--------|
| `.github/workflows/ci.yml` | Detection step, iOS/Android source flags, `.cxx` cache |
| `ios/Podfile` | `post_install` hook for source build header paths |
| `ios/Podfile.lock` | Updated by pod install |
| `scripts/postinstall.sh` | OpenCL ICD clone + bob build for git installs |
| `tools/use-llama-rn-source.sh` | New toggle script |
| `package.json` | Points to git ref |
| `yarn.lock` | Updated |
| `android/gradle.properties` | Uncommented `rnllamaBuildFromSource` |

## Test plan

- [ ] Push branch and verify CI detects `llama_from_source=true`
- [ ] Verify iOS CI build sets `RNLLAMA_BUILD_FROM_SOURCE=1` and compiles from source
- [ ] Verify Android CI build sets gradle property and compiles from source
- [ ] Verify `tools/use-llama-rn-source.sh <version>` switches back to npm correctly
- [ ] Verify switching back to npm version results in clean CI (no source flags set)

Story: TASK-20260406-0901


a-ghorbani:
Closing in favour of a fresh PR — the useful part here is the CI wiring, but three things in this branch are now stale or wrong, and rebasing costs more than restarting.

**1. The detection premise no longer holds.** The gate is:

```bash
if grep '"llama.rn"' package.json | grep -q '#'; then llama_from_source=true
```

i.e. "a git ref means source build, an npm version means prebuilts." That stopped being true with `llama.rn@0.13.0-rc.0`, which ships `rnllamaBuildFromSource=true` in its **own** `android/gradle.properties` — so an ordinary npm version now builds from source anyway. Run against `main` today this step would report `llama_from_source=false` while Android is in fact compiling from source. The replacement needs to key off the effective build mode, not the dependency spec.

**2. `package.json` / `ios/Podfile.lock` changes are obsolete** — they pin `0.12.0-rc.4` and a temporary crash-fix commit; `main` is on `0.13.0-rc.0`. Those files should not be in the new PR at all; the toggle belongs in tooling, not a committed pin.

**3. It predates the variant-allowlist knob.** `0.13.0-rc.0` added `RNLLAMA_ANDROID_VARIANTS` (`ORG_GRADLE_PROJECT_rnllamaVariants`). Without it a source build compiles the full tree for all seven arm64 variants; upstream's own CI builds three. Any from-source CI work should use it or build times will be unreasonable.

The new PR should also provision the Hexagon SDK, which this branch does not — see #858, where the current from-source build silently drops the NPU backend from released Android builds.

The approach and the CI structure here are still the right shape and will carry over.

PR #733, opened 2026-05-21
[Feat]: non-touch autostart deep link for bench E2E matrix

a-ghorbani:
## Why

The bench E2E matrix is driven entirely over adb except one step: starting the run, which required a UI tap on **RUN BENCHMARK MATRIX**. That tap goes through Android's synthetic-input layer (`adb shell input tap`), and on **HyperOS** (POCO F7 Ultra) and **MediaTek** (POCO X7 Pro) the OS silently drops the injected tap — `onPress` never fires and the matrix never starts. A real finger tap works, so the wiring is correct; the OS just rejects injected touches. The existing fallback chain (swipe → monkey → long-hold → off-center taps) didn't reliably break through, so every config on every device needed a manual human tap (~10–15 per PR validation), blocking unattended runs.

## What

Adds a non-touch trigger that invokes the **exact same** start path the button does. The E2E-only deep link now honors an `autostart` query param:

```
pocketpal://e2e/benchmark?autostart=1
```

When present, the `BenchmarkRunnerScreen` calls its existing `onRun` once after mount, loading the already-pushed `bench-config.json` — same code path, same single-flight gating, **no change to what is measured**.

- `parseBenchmarkAutostart` — pure query-only helper (truthy allowlist `1`/`true` case-insensitive; returns `false` on anything else or any parse error). Does not touch routing; `isBenchmarkRunnerUrl` remains the sole routing gate.
- Both deep-link delivery sites (`useDeepLinking` Linking effect + `dispatchAutomationDeepLink`) thread `{autostart}` from the raw URL via that one helper.
- Screen reads `route.params.autostart` and fires `onRun` once per mount via a `useRef` latch (independent of MobX-observer re-renders); single-flight remains the backstop.
- WDIO spec/helper updated: `deepLinkLaunch` sends `?autostart=1` and the spec drops the `.click()` (and the now-dead button wait), polling `bench-runner-screen-status` directly.

No native changes — the `pocketpal://` scheme is already registered in the e2e `AndroidManifest.xml`. No new DCE marker introduced, so the prod bundle-grep contract is unchanged.

## Testing

**Unit:** full suite green — 2268 passed / 2 skipped, global coverage gate PASS. Covers the canonical scenarios: autostart starts a run, bare URL stays `idle` (regression guard), autostart+tap race yields one run, `autostart=0`/garbage → false, and iOS/Android helper parity.

**On-device (the two devices that drop injected taps):** built the e2e APK, pushed a single-cell config, cold-launched the deep link with **zero taps**:

| Device | `pocketpal://e2e/benchmark` (bare) | `...?autostart=1` (no tap) |
| --- | --- | --- |
| POCO F7 Ultra (HyperOS) | `idle` ✓ | `running → complete` ✓ |
| POCO X7 Pro (MediaTek) | `idle` ✓ | `running → complete` ✓ |

The matrix started and ran to completion with no human touch on both devices; the bare URL correctly stays idle.

---

### DIFF
--- a/e2e/helpers/bench-runner.ts
+++ b/e2e/helpers/bench-runner.ts
@@ -71,8 +71,13 @@ export function pushConfig(
 }
 
 export async function deepLinkLaunch(): Promise<void> {
+  // `?autostart=1` makes the runner screen self-start its matrix run with no
+  // injected tap. This is the fix for HyperOS / MediaTek devices, where the
+  // OS silently drops `adb shell input tap` (and WDIO's analogous .click()),
+  // so a tap-driven start never fired onPress. The screen now invokes the
+  // exact same start handler the button does.
   await driver.execute('mobile: deepLink', {
-    url: 'pocketpal://e2e/benchmark',
+    url: 'pocketpal://e2e/benchmark?autostart=1',
     package: PACKAGE,
   });
 }
--- a/e2e/specs/benchmark-matrix.spec.ts
+++ b/e2e/specs/benchmark-matrix.spec.ts
@@ -65,20 +65,49 @@ describe('Benchmark Matrix', () => {
     const status = await driver.$(byTestId('bench-runner-screen-status'));
     await status.waitForDisplayed({timeout: 30_000});
 
+    const readStatus = async (): Promise<string | null> =>
+      (await status.getAttribute('content-desc').catch(() => null)) ??
+      (await status.getText().catch(() => null));
+
+    // Fast-fail diagnostic: with autostart=1 the screen must leave `idle`
+    // within seconds of mount. If it doesn't, the autostart param never
+    // reached the screen — fail in 30s with that specific signal instead
+    // of burning the full BENCH_MAX_WAIT_MIN on a silent timeout.
+    let lastObserved: string | null = await readStatus();
+    try {
+      await browser.waitUntil(
+        async () => {
+          lastObserved = await readStatus();
+          return typeof lastObserved === 'string' && lastObserved !== 'idle';
+        },
+        {timeout: 30_000, interval: 1_000},
+      );
+    } catch {
+      // Rethrow with the actual last-observed status interpolated. (WDIO's
+      // own timeoutMsg is a static string built before the wait starts, so
+      // it cannot include the value of `lastObserved` at timeout time.)
+      throw new Error(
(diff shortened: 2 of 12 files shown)


a-ghorbani:
Round-1 review folded in across 3 atomic commits (plus an arch-doc update in the control-plane repo). Full suite still green: 2268 passed / 2 skipped / 160 suites, coverage gate PASS.

### Concern 1 — `useRoute()` try/catch — **addressed** (`36c058a`)

Dropped the try/catch and the `__autostart` test seam. `useRoute()` is now called unconditionally; tests mock `@react-navigation/native` at the file level — the codebase's existing per-file pattern (same as `SquarePalCard`, `ModelCard`, `ModelNotLoadedMessage`, `useDeepLinking` test files). The autostart suite drives the trigger by setting `route.params` on the mock, which is exactly what the two deep-link delivery sites do in production via `parseBenchmarkAutostart`.

### Concern 2 — protocol in `navigationConstants.ts` — **addressed by extending scope** (`6b4d567`)

Moved the URL prefix, matcher, and autostart parser into a new `src/__automation__/benchmarkRoute.ts`. The reviewer's point was right at the boundary level — but moving only `parseBenchmarkAutostart` while `BENCHMARK_RUNNER_URL_PREFIX` and `isBenchmarkRunnerUrl` stayed behind would have been less consistent, not more. All three protocol-surface symbols moved together; `ROUTES.BENCHMARK_RUNNER` stays in the `ROUTES` enum because it's just a screen identifier, not protocol.

The two prod-reachable consumers (`useDeepLinking` and `App.tsx`) are already allow-listed in the `.eslintrc` `no-restricted-imports` rule as "the two legitimate mount points" — the move uses that escape hatch for exactly what it was carved out for.

Tests moved with the symbols (`src/utils/__tests__/navigationConstants.test.ts` → `src/__automation__/__tests__/benchmarkRoute.test.ts`). Architecture flow doc updated in the same change set.

### Concern 3 — autostart timeout diagnostics — **addressed** (`b76a1d0`)

Added a 30-second "left idle" probe right after `waitForDisplayed`. If autostart fails to fire, the spec now fails in 30 seconds with a specific message pointing at the deep-link plumbing / once-per-mount effect — instead of burning the full `BENCH_MAX_WAIT_MIN` on a silent timeout. Also interpolated the last observed status into both timeout error messages so operators see the actual stuck state.

---

PR #709, opened 2026-05-03
feat(chat): introduce structured assistant turns for tool calling

a-ghorbani:
## Summary

Replaces PR #705's tool-calling shape (`Text` + metadata bag) with a typed `MessageType.AssistantTurn` carrying `steps: AgentStep[]`, drives the agent loop from a pure async-iterator runner, and reduces events to UI state via a hand-rolled reducer. Renderer shows N visual blocks per turn within one FlatList row (preserves recycling). Long-press menu, copy / edit / tryAgain, and persistence all updated to handle the new type.

## Branching strategy

Branched from `1c1ecec` (PR #705's tip), **not** from `main`. PR #705's wrong on-disk shape never reaches main. **This PR supersedes #705** — close #705 with a redirect comment when this is merged. The abandoned `feature/TASK-20260415-1030` worktree (which tried to bolt multi-turn onto the wrong shape and ballooned the hook to 757 lines) will be removed via `./tools/remove-worktree.sh TASK-20260415-1030 --yes` after merge.

## Why

PR #705 (PACT talents) introduced tool-call state via `metadata.toolCalls / toolResults / completionResult / partialCompletionResult / isHtmlPreview / htmlContent` on a `Text` message. That shape doesn't model multi-turn dispatches (each turn is `[step₀, step₁, ..., stepN]` of `text + tool_calls + tool_outcomes`), conflates persistence with in-memory state, and pushes loop control into the hook. The data shape needs to be fixed first; the loop comes naturally on top.

## What's in this PR

- New message type: `MessageType.AssistantTurn` with `steps: AgentStep[]` top-level
- New service: `src/services/agent/` — pure runner (`AsyncIterable<AgentEvent>`), pure reducer (`(state, event) => state`), per-context trigger marker cache
- WatermelonDB persistence: `metadata.steps` is the persistence-only carrier; `toMessageObject()` lifts to top-level on read; `metadata.steps` is never read by hook / runner / reducer / renderer / selectors
- API conversion: `convertToChatMessages` walks `steps[]`; `derivedText()` is the single source of truth for text projection
- Renderer: ChatView computes `isActiveRun` once, passes down; Message renders one FlatList row per turn with N visual blocks (preserves FlatList recycling)
- Interaction: long-press menu, copy, edit, tryAgain, tryAgainWith all widened for `assistant_turn`
[...]

PR #736, opened 2026-05-23
fix(chat): tighten WebView JS surface + interrupted turn UX

a-ghorbani:
Follow-up to #709 (already merged). Addresses two pieces of post-merge review feedback that landed too late to make the squash.

## Summary

- **WebView lockdown (HtmlPreviewBubble).** The in-row `render_html` preview now sets `javaScriptEnabled={false}`, so scrolling chat history cannot trigger model-generated `while(true)` / `setInterval` CPU drain. The user-activated fullscreen modal keeps JS enabled and auto-closes via `useIsFocused` when the chat screen blurs, so leaving the chat unmounts the JS-running WebView instead of letting it burn cycles off-screen. CSP `default-src 'none'` continues to block all network exfiltration.

- **Failure UX (useChatSession + AssistantTurnFooter).** When the native tool-call parser rejects a truncated JSON args string (typically the model running out of context mid-args on a long `render_html` payload), the error is now routed through the turn's own metadata (`interrupted`, `truncationLikely`) instead of being dumped as a multi-KB native error string in a chat system message. The footer surfaces "Interrupted" or "Cut off — likely context full". When there's no partial content on the turn, a friendly `chat.toolCallTruncated` message stands in for the raw error.

- **L10n.** Adds `bubble.interrupted`, `bubble.truncated`, `chat.toolCallTruncated`.

## Why now (context)

Reviewer report for #709 flagged the WebView JS surface as a CONCERN: model-generated HTML can pin CPU just by being visible. Original spike comment explicitly accepted that residual risk — this PR closes it for the common case (chat history scrolling) while preserving the games-/interactive-HTML use case via the fullscreen modal.

The interrupted-turn UX fix is unrelated origin but discovered while testing the WebView change: when a tool-call fails to parse natively, the raw `[json.exception.parse_error.101]` dump that previously landed in chat was awful UX. Surfacing the failure on the turn itself is the right resting place.

## Reviewer note

Because #709 was squash-merged, the PR's **Commits** tab lists all 70 original commits even though only the last commit (`ba688e6`) introduces new changes against main. The **Files changed** tab is the authoritative view — exactly 8 files, +216 / -37.

## Test plan

- [x] `yarn typecheck` — clean
- [x] `yarn test` on affected suites: `HtmlPreviewBubble`, `AssistantTurnFooter`, `useChatSession.test.ts`, `useChatSession.assistantTurn.test.ts`, `locales`, `talents/*` — 353 passing, 0 failures
- [x] Manual: opened in-row HTML preview, confirmed CSS/layout still renders without JS; tap to expand → JS animations run in fullscreen; swipe to drawer → modal auto-dismisses, JS context unmounts
- [ ] Manual: reproduce a "context full mid-tool-call" failure and verify the footer shows "Cut off — likely context full" with no chat system-message dump


a-ghorbani:
Superseded — GitHub's merge-base computation against the squash-merged #709 inflated the diff to 15k+/1.2k across 165 files. Reopening from a clean cherry-pick branch.

PR #759, opened 2026-06-03
feat: add internet search functionality and related settings to ChatI…

kanhiayamishra152:
…nput

## Description

Fixes # (issue number)

## Platform Affected

- [ ] iOS
- [x] Android

## Checklist

  - [ ] iOS Simulator/Device
  - [x] Android Emulator/Device


a-ghorbani:
Thanks! internet search is something we want. The challenge: PocketPal already has a model-driven tool/talent system (src/services/talents + src/services/agent), and a Pal opts into tools via pact.talents. 

Web search should be a new TalentEngine registered there so the model decides when/how to search, rather than a global toggle that pre-fetches on every message. Separately, the PR currently (a) deletes the talent/agent imports while keeping their usages so it doesn't compile, (b) removes the pal-override settings resolver and the tool-compat banner, and (c) rewrites the v3→v4 migration — those need to be reverted; a new field should be a v5 migration. Also, api.duckduckgo.com is the Instant Answer API, not general web search, so it'll return nothing for most queries ... we'd need a real search backend.

a-ghorbani:
Closing as #808 supersedes this one.

PR #763, opened 2026-06-05
[Feat]: Warn when a chat is near or at the context limit, with recovery

a-ghorbani:
## Summary

Long conversations and tool-heavy talents (notably `render_html`) silently overflow `n_ctx`, producing truncated replies or "tool call cut off" errors the user can't diagnose. This adds a visible context-limit banner in the chat surface — now with a fullness meter and a context-size slider sheet — plus recovery actions.

## What changed

- **Near-limit warning + fullness meter** — when a finished turn used ~80%+ of the loaded `n_ctx`, a soft, dismissable "getting tight" banner appears with a fullness meter (`used / n_ctx`, clamped) and a percent readout. The trigger is purely ratio-based and talent-agnostic.
- **Sticky full banner** — when a turn was truncated or hit the context limit, a stronger sticky banner appears (also with the meter) and persists until a later turn drops back below the limit. Copy escalates after repeated full turns.
- **Context-size slider sheet** — the "More room" CTA opens a sheet where the user picks a larger context size on a slider over a context ladder, capped at the model's trained context length. Each stop shows a capacity readout (~words · ≈RAM) and a three-zone fit chip — **Fits / Tight / Won't fit** — classified with the same memory estimator the model-load path uses. Confirm is disabled on a "Won't fit" stop. An Advanced disclosure shows the raw `from → to` tokens and model max.
- **OOM-safe, never a dead-end** — the banner's increase CTA appears only when at least one larger tier fits the device. When nothing larger fits, the sheet (still reachable from the pal-load hint) hides confirm and offers a New-chat affordance instead of a permanently disabled button.
- **Recovery reload** — confirm reloads the model at the chosen size; on failure it restores the previous `n_ctx` and re-initialises the model so it ends up loaded, not just the setting restored. Chat history is preserved (messages live in the store, not the native context).
- **Remote (PalsHub) models** — a hedged "reply may be cut off" advisory on weak truncation signals, with no meter and no increase-context option (PocketPal does not own a remote server's `n_ctx`).
- **Quiet when inactionable** — no banner when no model is loaded, when dismissed for the draft; the existing HTML soft-cap is folded into the same single banner slot (context variants take precedence).
- **Talent-aware copy + pre-load hint** — an optional `recommendedContextTokens` hint on `TalentEngine` (declared by `render_html`) drives a one-shot snackbar when a heavy-talent pal loads into a chat smaller than its recommendation; the snackbar's "More room" action opens the slider sheet. It never moves the trigger threshold.
- **Footer de-duplication** — the assistant-turn footer drops its "cut off" wording on the turn that drives the sticky banner.

## Implementation notes

- A normalised completion snapshot is written once per finished turn at both completion boundaries (`run_finished` and the abort-with-partial path), mirrored onto the message metadata and the session store in one action. Banner state is MobX-ephemeral (no DB column, no migration).
- The banner variant is chosen by a pure resolver (`resolveBannerVariant`) that also emits the meter `ratio`; it has no JSX, MobX writes, or store coupling.
- The sheet owns target selection; a pure, caller-injected three-zone fit classifier (`makeFitStatusFor`) keeps the memory math store-free. Slider pick / advanced / device-total are local component state.
- New English strings only; other locales follow via Weblate.

## Verification

- Lint, `tsc --noEmit`, and the full Jest suite pass; l10n validation passes.
- Unit/render coverage for the resolver (`ratio` + variant precedence), the meter and CTA gating on the banner row, and the slider sheet (capacity, fit zones, won't-fit-disables-confirm, no-fit confirm-hidden + New-chat, arbitrary current `n_ctx`, restore-on-failure re-init).

## Remaining before merge

- Visual capture across the 8 integrated locales (incl. RTL `he`) on a narrow device (iPhone SE): meter banner variants (warning + full), the slider sheet with all three fit-chip zones, and the no-fit sheet state.
- Manual check of the live increase-context reload and restore-on-failure path.


a-ghorbani:
Follow-up fixes applied after the initial review pass (commits e2b99f8..5618db7):

- **Remote hedged advisory now renders.** The resolver previously gated the remote-hedged variant behind a known runtime `n_ctx`, which remote models never set — so the advisory was unreachable. It now gates on a loaded model only; the n_ctx-reading variants (full/warning) still require a known `n_ctx`. Added resolver + `BannerRow` tests for the remote case.
- **Failed/cancelled context reload now re-loads the model.** `IncreaseContextSheet` previously restored only the `n_ctx` setting after a failed reload, leaving the model unloaded while the copy claimed it was restored. The catch now re-inits at the prior `n_ctx` (with a nested guard if that also fails) and keeps the New-chat CTA reachable.
- **Localized talent name in the banner.** The full banner showed the raw engine key (`render_html`) instead of its localized display name ("HTML preview"); it now maps through the existing talent-name table with a safe fallback.
- **Narrow-width / RTL banner layout.** The warning and remote-hedged row banners now let their text shrink/wrap (`flex:1`/`flexShrink:1`) instead of clipping or pushing Dismiss off-screen on iPhone SE and in RTL.

Verification: `tsc --noEmit`, lint, and the full Jest suite pass (220 suites, 3222 tests); l10n validation passes.

Still remaining before merge (unchanged): visual capture across the 8 integrated locales incl. RTL on a narrow device, and a manual pass of the live increase-context reload / restore-on-failure path.

a-ghorbani:
UI refresh + review fixes (commits 8f5fc4b..52fef91):

The chat-surface warning now matches the design prototype, on top of the existing data layer:
- **Fullness meter** on the warning/full banners (driven by the real used/n_ctx ratio), pill actions, and per-state styling.
- **Context-size sheet** is now a slider over a token ladder (capped at the model's trained context length) with a Fits / Tight / Won't-fit chip, a plain-language capacity readout (~words · ≈RAM), a status line, and an Advanced disclosure for raw tokens. When the device can't fit any larger size, the sheet offers New chat instead (no dead-end).
- **Pal-load hint** snackbar gained a "More room" action.

Review fixes in this round:
- Fixed a layout bug where the warning banner's fullness meter collapsed to zero width (it now stacks and renders full-width).
- The banner's "more room" affordance is now hidden when the chat is already at the model's maximum context, so it can't open a no-op resize.
- Added screen-reader label/value to the size slider (announces the size, not a raw index) and hid the decorative meter from assistive tech.
- Doc/architecture notes reconciled with the shipped behaviour.

Verification: tsc, lint, and the full Jest suite pass (221 suites, 3241 tests); l10n validation passes. No native changes.

Remaining before merge: visual capture across the 8 integrated locales (incl. RTL) on a narrow device (iPhone SE) for the meter banner, the slider sheet (fit-chip zones + no-fit state), and a manual pass of the live increase-context reload / restore-on-failure path.

a-ghorbani:
Follow-up polish + fixes from manual testing (commits since the UI-parity round):

- **Banner visuals reworked** — replaced the saturated background + heavy filled/outlined buttons with a soft accent-tinted wash and plain right-aligned text actions, so the recovery buttons fit the banner instead of clashing.
- **Full banner is now dismissable** — added a Dismiss action (per-draft; it reappears on the next turn if still full). Previously sticky-only.
- **Banner clears on edit/regenerate** — editing an earlier message (or regenerating) shortens the context, so the stale banner now drops immediately and re-evaluates on the next reply, instead of waiting for a full turn.
- **Increase-sheet button layout** — the confirm/cancel buttons no longer cram to the left (they fill the action row).
- **Reload result snackbar** now auto-dismisses (it previously stayed on screen because the timer didn't re-arm across the reloading→result transition).
- **Settings reactivity** — the Settings screen now re-syncs the displayed context size when the global n_ctx changes elsewhere (e.g. the banner's increase-context flow); it previously showed a stale value until remount.
- **E2E** — added `e2e/specs/features/context-banner.spec.ts` covering the reachable banner lifecycle (no banner on fresh chat, banner+meter after a reply overflows a tiny context, open/cancel the increase sheet, dismiss, New chat from the full banner).

Verification: tsc clean, lint clean, full Jest suite passes (221 suites, 3243 tests). No native changes.

Still remaining before merge: visual capture across the 8 integrated locales (incl. RTL) on a narrow device for the new banner look + the increase sheet, and a device run of the new E2E spec.

PR #774, opened 2026-06-12
Feat: Show model name in chat (#755) + HF-Mirror support (#773)

### LINKED ISSUE #753: [Bug]: URLs not distinguishable from regular text
**Describe the bug**
When a model response includes a URL (in markdown format typically), it is rendered exactly the same as other text on the screen.  Please add the ability to configure the color/style of links.

Using app version 1.14.0 (137).

### LINKED ISSUE #755: [General]: record/show model used for chat responses
**Issue description**
When a previously created chat session is re-opened using the hamburger menu, it does not indicate for a given response what model was used to generate that response. This information can be helpful to know, so it would be nice if it were displayed below each response, or if there were a way to query it.

### LINKED ISSUE #773: [Feat]: HF-Mirror support request（HF-Mirror支持请求）
**Description**
Access to Hugging Face is blocked or difficult in mainland China, but HF‑Mirror can effectively alleviate this issue.
中国大陆地区无法或难以访问HuggingFace，HF-Mirror可有效缓解这个问题

**Use Case**
It will be used when downloading models in mainland China :)
在中国大陆下载模型时会用到:)

ghshhf:
### Fix #755: Show model name in assistant turn footer
- Add `modelId` field to `AssistantTurn` type
- Save model ID when creating AI response
- Display model name in `AssistantTurnFooter`

### Fix #773: Add HF-Mirror support for Chinese users
- Add `useHfMirror` state to `HFStore`
- Support dynamic domain selection (huggingface.co / hf-mirror.com)
- Add settings switch for HF-Mirror
- Add localization strings (en, zh)

### Fix #753: Style URLs to be distinguishable (included)
- Add `<a>` tag styling in MarkdownView

## Testing

- [ ] Model name displays correctly in assistant turn footer
- [ ] HF-Mirror switch works in Settings
- [ ] URL styling renders correctly

## Notes

- `modelId` is optional for backward compatibility
- HF-Mirror uses `hf-mirror.com` as the mirror domain

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,152 @@
+## 🎯 Pull Request: #755 + #773 功能实现
+
+### 📋 关联的 Issue
+- Fix #755: Show model name in assistant turn footer
+- Fix #773: HF-Mirror support request (HF-Mirror支持请求)
+- (附带) Fix #753: URL style improvement
+
+---
+
+### 🚀 实现的功能
+
+#### 1. **Fix #755: 在 AI 回复中显示模型名称**
+**问题**: 用户重新打开旧对话时，不知道每条回复是用哪个模型生成的。
+
+**解决方案**:
+- ✅ 修改 `AssistantTurn` 类型，添加 `modelId` 字段
+- ✅ 在创建 AI 回复时保存当前模型 ID
+- ✅ 在回复底部显示模型名称（使用主题主色）
+
+**修改的文件**:
+- `src/utils/types.ts`
+- `src/hooks/useChatSession.ts`
+- `src/components/AssistantTurnFooter/AssistantTurnFooter.tsx`
+- `src/components/AssistantTurnFooter/styles.ts`
+
+---
+
+#### 2. **Fix #773: 添加 HF-Mirror 支持（中国大陆用户）**
+**问题**: 中国大陆用户无法访问 Hugging Face（被墙），导致无法下载模型。
+
+**解决方案**:
+- ✅ 在 `urls.ts` 中支持动态域名切换
+- ✅ 添加 `useHfMirror` 状态到 `HFStore`
+- ✅ 在设置页面添加开关（"Use HF-Mirror (China)"）
+- ✅ 添加本地化字符串（英文、中文）
+
+**修改的文件**:
+- `src/config/urls.ts`
+- `src/store/HFStore.ts`
+- `src/screens/SettingsScreen/SettingsScreen.tsx`
+- `src/locales/en.json`
+- `src/locales/zh.json`
+
+---
+
+#### 3. **(附带) Fix #753: 改进 URL 样式**
+**问题**: AI 回复中的 URL 和普通文本样式一样，无法区分。
+
+**解决方案**:
+- ✅ 为 `<a>` 标签添加样式（主题主色 + 下划线）
+
+**修改的文件**:
+- `src/components/MarkdownView/styles.ts`
+
+---
+
+### 📸 效果预览
+
+#### Fix #755: 显示模型名称
+```
+User: Hello!
+Assistant: Hello! How can I help you today?
+             [Model Name]  [42 ms/tok]
+```
+
+#### Fix #773: HF-Mirror 开关
+设置页面 → Model Initialization Settings → "Use HF-Mirror (China)"
+- 开关：Off [切换到 On]
+- 描述：Use hf-mirror.com instead of huggingface.co...
+
+---
+
+### 🧪 测试计划
+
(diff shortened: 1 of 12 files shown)


a-ghorbani (changes requested):
## Review summary — Request changes

Thanks for the contribution! This PR bundles three unrelated features (#755 model name in footer, #773 HF-Mirror, #753 link styling). A few blocking issues need to be resolved, and I'd recommend splitting the three features into separate PRs so the cosmetic fix isn't held up by the others.

### Blockers

**1. The PR does not compile — `HF_DOMAIN` export was removed but is still imported.**
`src/config/urls.ts` renames `HF_DOMAIN` → `HF_DOMAIN_OFFICIAL` (+ `getHFDomain()`), but two files still import the old symbol:
- `src/utils/hfResolve.ts:2` (used at `:82`, `:102`)
- `src/store/PalStore.ts:22` (used at `:359`)

This is a TypeScript error (`HF_DOMAIN` has no exported member); if it reached a bundle it would produce `undefined/<repoId>` URLs. Please migrate both call sites to `getHFDomain()` (which also makes them mirror-aware).

**2. `modelId` is never persisted — the model name disappears on reload (#755).**
`modelId` is set in-memory in `src/hooks/useChatSession.ts`, but:
- the write path in `src/repositories/ChatSessionRepository.ts` only folds `steps`/`authorData`/`imageUris` into the `metadata` column — `modelId` is dropped;
- `src/database/models/Message.ts` (`toMessageObject()`) reconstructs `assistant_turn` rows without `modelId`.

So the footer shows the model name only for turns generated in the current session; it vanishes for all historical turns after switching sessions or restarting the app. Fix: persist `modelId` inside `metadata` on write and lift it back in `toMessageObject()`, mirroring the existing `steps` pattern. No schema migration is needed (it rides the existing JSON `metadata` column; older rows stay safe since `modelId` is optional).

**3. HF token is forwarded to the third-party mirror (#773).**
With both "Use HF Token" and "Use HF-Mirror" enabled, the stored HF Bearer token (which grants access to private/gated repos) is sent in the `Authorization` header to `hf-mirror.com`. The token is attached domain-agnostically at every HF call site (`src/api/hf.ts`, `src/services/downloads/DownloadManager.ts`, `src/utils/hf.ts`) with no domain guard. Please suppress the `Authorization` header for any non-`huggingface.co` domain, and note in the setting copy that private/gated models aren't supported via the mirror.

**4. Don't edit `src/locales/zh.json` directly.**
Only `src/locales/en.json` should be hand-edited — other locales are managed through Weblate and will be overwritten (see `CONTRIBUTING.md`). Please revert the `zh.json` change and keep the new keys in `en.json` only; the Chinese translation can go through Weblate.

### Concerns

- **Persisted mirror setting doesn't apply on cold start.** `HFStore`'s constructor calls `setHFDomain(this.useHfMirror)` synchronously with the default `false`, but `mobx-persist-store` rehydrates `useHfMirror` asynchronously afterward and nothing re-applies the domain. A persisted `true` reverts to `huggingface.co` until the user re-toggles — exactly the wrong failure mode for users who can't reach Hugging Face. Re-apply on hydration (e.g. a `reaction(() => this.useHfMirror, setHFDomain, {fireImmediately: true})`), or better, drop the module-level mutable domain.
- **Module-level mutable `currentHFDomain` in `src/config/urls.ts`** creates a dual source of truth with `HFStore`. Consider deriving the domain from the store (read at call time) instead of imperatively pushing it into the config module — this also removes the root cause of the two issues above.
- **No tests** were added for any of the three features (footer model name, `modelId` round-trip, mirror toggle + persistence, dynamic URL domain).
- **Mirror-served model binaries aren't integrity-checked.** Download URLs are redirected to the mirror and written without checksum/signature verification. This is inherent to mirror support, but worth a product decision on scope and a user-facing note.

### Minor

- Remove the issue-tag / narration comments (`// FIX #755`, `// Fix #753`, `// Look up model by ID …`, the `Added for #755` doc comment) — comments should describe current state, not reference issues.
- `PR_DESCRIPTION.md` and `PUSH_AND_PR_GUIDE.md` shouldn't be committed to the repo.
- `src/components/MarkdownView/styles.ts` hardcodes `fontSize: 16` on the `a` tag — let it inherit the body size instead.
- The footer model name uses `theme.colors.primary` + `fontWeight: '500'` while sibling footer text uses `textSecondary`; consider matching for a calmer footer.
- "Use HF-Mirror (China)" — consider dropping the region parenthetical; the description already explains the connectivity condition.

Verification note: native build not required (no native surfaces touched). A full typecheck/test run wasn't executed, but the compile break, the persistence gap, and the token forwarding are all confirmable directly from the source above.

PR #793, opened 2026-06-21
[Feat]: Redesign — My Pals screen + Create/modify pal General/Generation tabs

a-ghorbani:
## Summary

Redesign slice for the Pals area (reskin + structural UX), wired to the Phase-2 design system and Phase-1 tokens.

**My Pals screen** (`PalsScreen`): reskinned to a local-only "My Pals" surface — serif header (back + title + "+ Create Pal"), segmented **Downloaded | Created by me** tabs, single-column pal cards (avatar, name, subtitle, category chips, "…" overflow → edit / share / delete). The unified screen's hub-discovery chrome (auth bar, search, filter chips, bottom action bar, remote library/discovery rendering) is dropped here — the Explore tab now owns discovery.

**Create / modify pal form** (`PalSheet`): the form is now a tabbed **General | Generation** sheet. The former standalone generation-settings sheet is folded into the Generation tab as a pure relocation — the persistence path is unchanged. General keeps Name / Description / Model / schema-driven fields (World/Location for roleplay, Capture Interval for video) / System Prompt / Theme / Talent / Greeting.

### Write path unchanged
`completionSettings`, `pact`, and `greeting` are still written **only** through `PalSheet.onSubmit → palStore.create/updatePal`. The Generation tab's Reset / Clear menu mutates in-form state only and persists on Save — never eagerly. Tests assert both directions (persist on Save; no store write on Reset/Clear alone).

### testID re-home note
The live create-path testID `bottom-action-add` migrates from the old bottom action bar onto the new header "+ Create Pal" affordance (single mounted node — asserted in tests). Form-internal and card testIDs are frozen. Discovery-only testIDs are dropped together with the surfaces they belonged to.

### Tests
- Full suite green: **250 suites, 3743 passed** (2 skipped); coverage 75% statements / 70% branches (req 60%).
- `yarn lint` and `yarn tsc --noEmit` clean.
- New coverage: My Pals tab routing, create-type chooser labels, per-tab empty states, single-`bottom-action-add` collision guard, roleplay create persistence, Reset-to-system / Clear single-writer guards, schema-driven Capture Interval gating.
- No native changes (no package.json / ios / android / Podfile / gradle touched).

### Owed before flip
- Device visual captures (light / dark / forced-RTL on a tab switch / non-Latin) for My Pals (both tabs) and the tabbed form — PM-serialized across simulators, owed to the consolidated pre-flip device pass.
- Architecture-doc delta (`pals-and-talents.md`) applied on-disk in the control-plane repo; PM to commit it there (this app PR cannot carry it).


a-ghorbani:
## Review fix round 1

Addressed the review findings:

- **Validation guard restored** — the pal form's Generation tab now re-runs the shared completion-settings validation on Save and blocks persist (alert + early return) on out-of-range numbers or non-numeric input in numeric fields, matching the chat settings path. Added a regression test asserting invalid settings are neither created nor updated.
- **Dark back chevron** — normalized the chevron SVG fill so the color prop applies, and pass explicit size + fill (not stroke) at the call site so it renders in dark mode.
- **Bulk import / export-all** — re-homed into a header overflow menu (the custom header had dropped the only entry point), reusing the existing import/export utilities.
- **Accessibility** — back button now announces "Go back"; the tappable pal card is a button with an edit hint; added hit slop / minimum touch targets to the back, create, and overflow controls.
- **Cleanup** — stopped falling back to the raw system prompt for the card subtitle; removed dead styles, stale source strings, and orphaned discovery components/barrel exports.

Full test suite green (249 suites, 3727 passed); typecheck and lint clean.

Issue #355, opened 2025-06-26
[Bug]: In the new version, some models cannot converse normally, and the visual model fails to recognize images.


cxay666:
Describe the bug
In the new version, some models cannot converse normally, and the visual model fails to recognize images.
 
To Reproduce
 
1. Go to "hf" and download the SmolVLM2-500M-Instruct (Q8_0) model.
2. Use the SmolVLM2-500M-Instruct (Q8_0) model to answer questions.
3. Send "hi", and the model returns meaningless characters. When sending an image with a query, it displays "Completion failed: Failed to evaluate chunks".
 
Expected behavior
The model should answer questions normally and recognize image content.
 
Screenshots
(If applicable)

![Image](https://github.com/user-attachments/assets/56044860-f835-42d9-a2a2-9e456fd831a2)

 
Environment (please complete the following information):
 
- Device: Honor X30
- OS: Android 13
- App Version: 1.10.8
 
Additional context
No such issues occur when rolling back the app version to 1.10.7.

a-ghorbani:
@cxay666 thanks for reporting. 
which model throw `///////// ....`? was it also SmolVLM2-500M-Instruct (Q8_0)?

cxay666:
> [@cxay666](https://github.com/cxay666)感谢您的报告。 哪个型号抛出了`///////// ....`？也是 SmolVLM2-500M-Instruct (Q8_0) 吗？

Yes.

a-ghorbani:
Did this happen only once, or did you try reloading the model again to see if it was just a one-off issue?
I tested it on a Samsung A23, which has similar hardware, but I couldn't reproduce the error.

btw, did you install from apk or Play Store? 

<img width="458" alt="Image" src="https://github.com/user-attachments/assets/e90e7391-725b-42b3-b384-112e7dd0e4c1" />

<img width="794" alt="Image" src="https://github.com/user-attachments/assets/da77ce7e-b164-4ea9-a1ad-568efa2bd21d" />

cxay666:
> Did this happen only once, or did you try reloading the model again to see if it was just a one-off issue? I tested it on a Samsung A23, which has similar hardware, but I couldn't reproduce the error.
> 
> btw, did you install from apk or Play Store?
> 
[...]

PR #805, opened 2026-06-24
[Feat]: Speculative decoding / MTP draft-model support

a-ghorbani:
## Summary

Adds MTP / speculative-decoding draft-model support, consuming the published **llama.rn@0.12.5** — the release that synced llama.cpp to b9769 and shipped the MTP / speculative draft-model work from llama.rn PR #355 — as a normal **prebuilt** registry dependency (standard `0.12.4 → 0.12.5` upgrade). No from-source build, no vendored tarball.

Two modes, both supported:
- **Embedded / hybrid MTP** — a self-contained model carries draft layers; `model_draft` left unset, speculative just enabled (MTP defaults: `flash_attn_type 'auto'`, draft cache `q8_0`).
- **Separate-draft pairing** — a target + a distinct small draft model loaded together; `model_draft` = the draft file path (paired defaults: `flash_attn_type 'off'`, draft cache `f16`, draft GPU layers 99).

Mode is **derived** from whether a paired draft resolves to a downloaded file — no persisted mode enum. All mode defaults are applied as `?? user-value` fallbacks, so an explicit user setting is never overwritten.

## What changed

- **llama.rn upgrade (prebuilt)**: bumped `0.12.4 → 0.12.5` in `package.json` / `yarn.lock` / `ios/Podfile.lock`. The published release ships the prebuilt assets (iOS `rnllama.xcframework`, Android jniLibs for arm64-v8a + x86_64), so the net native diff vs `main` is exactly those three dependency-lock files — a plain upgrade. No committed source tarball, no `rnllamaBuildFromSource`/`RNLLAMA_BUILD_FROM_SOURCE` flag, no Podfile header-path injection, no `post_install` `project.pbxproj` mutation, and no OpenCL ICD stub. The four native config files (`ios/Podfile`, `ios/PocketPal.xcodeproj/project.pbxproj`, `android/gradle.properties`, `.github/workflows/ci.yml`) are byte-identical to `main`; `pod install` leaves the tree clean.
- **Persistence**: `ContextInitParams` gains `speculativeEnabled`, `selectedDraftModelId`, and `spec_draft_*`; `CURRENT_CONTEXT_INIT_PARAMS_VERSION` is `2.3` with the feature defaulting OFF for upgraded records (no behaviour change for existing users). `selectedDraftModelId` is optional, so it needs no migration step or version bump.
- **Draft parse path**: a cross-repo `draft.{hf_repo,hf_filename}` block in the device rules, parsed with the **same** path-segment security guard as model/mmproj (rejects path separators and `.`/`..` traversal; GGUF-only; huggingface.co host). A malformed draft block drops only the draft, keeping the target.
- **ModelStore**: read-only `resolveDraftConfig` (pre-mutex) with a two-channel precedence — a per-target device-rules-authored draft wins, the global user-picked draft is the fallback; two-mode `initLlama` params; `activeDraftModelId` single-writer with all release resets mirroring the projection-model pairing; best-effort draft auto-download honouring the same precedence (a draft download failure never fails the target download). The pre-load memory check now sums the target, projection, **and the paired draft's own weights and KV cache** (sized by the target `n_ctx` and the draft cache type).
- **Settings**: a speculative master toggle, a **global draft-model picker** (downloaded non-projection models + a "None" entry), a **draft GPU-layers control**, and draft cache-type menus. The draft cache-type label shows the **effective** resolved default (f16 paired / q8_0 embedded); the cache menus are enabled when a separate draft is paired and disabled with an explanation for embedded MTP, matching the runtime contract.

## Verification

- Lint (0 errors), typecheck, and the full Jest suite (247 suites, 3780 passed / 2 skipped) all green.
- `pod install` leaves the working tree clean (no generated `project.pbxproj` mutation); the committed checkout builds as-is.
- iOS Release build: PASS (prebuilt `rnllama.xcframework`; AppIntents wrapper compiles).
- Android `assembleProdRelease`: PASS — the APK packages **both** arm64-v8a and x86_64 from the prebuilt jniLibs (no from-source ABI risk).

## Open items / follow-ups

- **Device verification owed** (physical iOS + Android, may run after this draft):
  - speculative no-op gate: a non-MTP model with speculative enabled and no draft loads with **no native error** and produces output (the only proof of the speculative no-op assumption), on iOS and Android.
  - visual captures: no-op load, paired-draft-active, paired-draft-missing, and the Settings advanced speculative section.


a-ghorbani:
Independent review round 2 (final) is complete — all round-1 findings are resolved and the one net-new finding has been fixed.

**Dependency simplification:** since round 1, llama.rn #355 merged and shipped as the published **v0.12.5** (prebuilt). This PR now consumes it as a standard registry dependency instead of building from source — the net native diff vs `main` is just `package.json`, `yarn.lock`, and `ios/Podfile.lock`. This removed the entire from-source build apparatus (no committed binaries, no `post_install` project mutation, no per-ABI build risk).

**Round-2 review:** 6/8 role reviews clean; the one CONCERN (a stale global draft selection persisting after the picked draft model was deleted) is fixed — the selection is now cleared on delete and the Settings UI derives its paired/embedded state from the *resolved* draft model, so a stale id is coherently treated as "None".

**Verification:** typecheck clean, lint 0 errors, full unit suite green (3784); `pod install` clean; iOS Release and Android `assembleProdRelease` (both ABIs) build on the prebuilt 0.12.5.

**Still owed before this leaves draft:** on-device verification on physical iOS + Android — load a non-MTP model with speculative decoding enabled (no separate draft) and confirm no native error, plus visual captures of the paired-active / paired-missing / Settings states. These need physical devices and could not be run headlessly.

a-ghorbani:
**V1 verification — done, via a new automated e2e spec (no physical device needed).**

Added `e2e/specs/features/speculative.spec.ts` (mirrors `thinking.spec.ts`) and ran it on the iOS simulator (`yarn e2e:ios --spec speculative --devices virtual-only`). Result: **1 passing, 0 failing.**

The V1 "speculative no-op" gate is now proven end-to-end on a virtual device:
- Enable Speculative Decoding in Settings → Advanced Settings (no draft paired)
- Load a non-MTP model (Qwen3-0.6B) → it loads and produces output with **no native error**: `"Hi there! How are you today? 😊"` at ~109 tokens/sec.

The spec also captures the enabled Settings state — which visually confirms the review fixes: the Draft cache type shows its **effective default** ("Q8_0") and is **disabled with a reason** ("Draft cache type applies only when a separate draft model is paired") in embedded mode, and the draft picker reads "None (embedded MTP)".

Note: this is a normal RN Settings + model-load surface, so it verifies on a **simulator/emulator** via the standard Appium pipeline — it is not an App-Intents/Shortcuts path and needs no physical device. (A small `testID` was added to the Advanced Settings accordion to make the section reliably targetable in e2e.)

a-ghorbani:
**MTP speculative decoding now actually engages — with durable telemetry.**

A correctness review found the original wiring was **inert**: the activation parameter (`spec_type`) was never sent to llama.rn, so speculation never ran and a paired draft never loaded — silently, with no error, in both modes. (Unit tests passed because they couldn't observe engagement.) This was fixed:

- **Capability-gated activation.** `resolveDraftConfig` now resolves to `paired | embedded | off` by actual MTP capability (target `nextn_predict_layers` for embedded; a width-matched MTP draft for paired), and emits `spec_type=draft-mtp` only when capable. A non-MTP model with speculative on resolves to `off` and never activates — which also avoids an uncatchable native abort that emitting `draft-mtp` on a non-MTP model would trigger. The JS width pre-check exactly mirrors the native assert.
- **Engagement telemetry (durable).** `draft_tokens`/`draft_tokens_accepted` are now mapped onto the completion result and shown in the chat turn footer (`draft: accepted/total (pct%)`) when speculation engaged. This makes engagement observable in the UI and serves as a permanent signal so a future regression to "inert" can't hide behind green tests.
- **Capability detection + badging**, and a build fix so the new pre-download header-detection dependency resolves in the React Native bundle.

**Verified on device (iOS simulator):** speculation engages — e.g. `draft: 43/117 (37%)` (and 83% acceptance on a longer run); off-safety (non-MTP) and crash-safety scenarios pass; e2e 3/3 green. Independent review: no blockers; robustness/test-coverage follow-ups addressed (dependency pinned + bundle guard, component tests for the new surfaces, advisory-note accuracy, Android bundle confirmed).

a-ghorbani:
## Visual evidence — draft cache gating + speculative generation rate






a-ghorbani:
## Visual evidence — complete set (supersedes earlier partial post)

















a-ghorbani:
Notes on the visual-evidence set above (iPhone 17 Pro simulator, iOS 26.0, release build):

- **01–04** — Speculative section with the toggle off, then on (draft-model picker, draft GPU layers, draft cache rows).
- **05** — draft pick that carries no draft layers → "This model carries no draft layers, so it can't be used as a draft model."
- **06** — speculative on with a non-capable active model → "…stays off for it with no effect."
- **07 vs 08** — the draft cache rows greyed out with the "applies only once speculative decoding actually runs" hint, versus live with their value labels once the pick is a usable draft. **09** shows the draft key-cache menu open, i.e. the row is interactive.
- **10 vs 11** — "Vision, Speculative (MTP) capabilities." on the MTP model card versus no capability line on the non-MTP card.
- **12** — **the MTP badge does NOT render** in the HuggingFace details view for an MTP repo. The remote header for that repo does report `qwen35.nextn_predict_layers = 1` (verified independently with `@huggingface/gguf` outside the app), so this is a real gap rather than a fixture problem: the remote probe calls `TextDecoder`, which Hermes does not provide, so `gguf()` throws and the capability check silently resolves to false. **13** is the non-MTP repo (correctly no badge).
- **14 vs 15** — the assistant-turn footer with `draft: 47/114 (41%)` on an MTP turn, versus a non-MTP turn with no draft segment.

Two states are not in the set:
- The `speculativeDraftModelIgnoredOverridden` note needs an active model whose device rule ships its own `draft`; no entry in the published device rules has one, so it is unreachable on device today.
- The embedded-MTP variant of the settings screen (ignored-note "isn't compatible" + cache rows in embedded mode) needs the MTP target loaded. The MTP fixture ships in a vision repo and the iOS simulator's Metal driver aborts the process while loading the projection model (`clip_model_loader::load_tensors` → `MTLSimDevice newBufferWithLength:`), so the target never becomes active there. The cache-rows-enabled state in **08/09** is the paired-draft resolution of the same rows.

a-ghorbani:
## Visual evidence — CURRENT set at 316182bd (supersedes all earlier posts; MTP badge now renders)














a-ghorbani:
## Visual evidence — separate-draft (assistant/mem-shared) MTP engages on device: draft 183/351 (52%) @ 12.29 tok/s, Pixel 9 CPU




a-ghorbani:
## Visual evidence — draft-only handling on the Models screen (Pixel 9, physical device, commit 8e230423)

A speculative draft (`mtp-gemma-4-E2B-it-Q8_0`) is now recognised as draft-only: Load is disabled with an inline explanation, the card carries the chat icon instead of the vision eye, and the reported size is the draft file alone (97.82 MB — previously 655.19 MB because the mmproj was wrongly paired in). The target model above it keeps a normal Load.



Expanded card: capabilities read **Speculative draft** only — the Vision tag and Vision toggle are gone. This state was reached from a device that had downloaded the draft *before* draft detection existed: the persisted vision classification is re-derived at store init, so existing installs heal without re-downloading.



a-ghorbani:
## Speculative e2e suite — recorded run at ace3532b (Android emulator)

Off-safety (non-MTP + speculative on → loads, outputs, no native error) and crash-safety (width-mismatched paired draft → process survives) both **pass** on the emulator. Three harness faults that had kept the whole suite red are fixed in `ace3532b`: a Settings navigation race, a download tap silently skipped for file cards below the fold, and the model card's vision toggle being unreachable by automation (it now carries a testID and a screen-reader label).

The embedded-engagement test is **environment-blocked on a 2 GB emulator**: the identical build engages on a Pixel 9 (42.7 tok/s embedded; paired `draft: 183/351 (52%)` @ 12.29 tok/s, posted above), but on the AVD `llama.rn` cannot allocate the MTP draft context for the hybrid Qwen3.5 — completion fails with `failed to create MTP draft context`, with GC pressure logged at the same instant, with or without the repo's mmproj attached. The paired spec additionally needs a physical device because its 2.84 GB target does not fit the AVD's free storage.

**Known limitation worth a follow-up issue:** on a low-RAM device, an MTP-capable model with speculative enabled can fail completion with this same hard error instead of degrading to non-speculative generation. The throw originates inside llama.rn's completion path (`rn-completion.cpp`), so the clean fix is an app-side retry without speculative on this specific error, or an upstream fallback.

a-ghorbani:
## Round-5 fixes landed — and the device run they demanded caught a real bug

All six round-5 findings are addressed at `97e4a3d7`:

- **Heal hardening**: a downloaded record whose GGUF header hasn't been read yet is no longer classified by filename alone (the header gets to veto first), and orphaned-mmproj cleanup is now a persisted queue drained on every launch — a kill, an in-flight download, or a failed delete can no longer strand the file. Shared projectors (still referenced by a downloaded vision model) are dropped from the queue, never deleted.
- **Error copy**: the low-RAM speculative failure now maps to actionable copy naming the speculative toggle instead of a raw native dump; deeper handling (auto-retry without speculative, estimator term) is tracked in #850.
- **Test/a11y/hygiene residue**: remaining scenario anchors renamed, the e2e vision-disable helper scopes its card query and scrolls before concluding absence, and the model card's vision switch carries a testID + screen-reader label.

**The important part:** running the engagement gate on a 12 GB Pixel 9 reproduced the `failed to create MTP draft context` failure previously written off as an emulator-RAM limitation — it wasn't. llama.cpp refuses a quantized draft V cache whenever flash attention resolves off, and `flash_attn_type='auto'` resolves off on Android CPU *inside* context init. Our embedded draft-cache default was `q8_0`, so **embedded MTP silently never engaged for default-settings Android users**. Fixed: the embedded draft cache defaults to f16 unless flash attention is explicitly forced on, and a quantized draft-V pick is clamped to f16 when flash attention is off. Regression-tested at the emission, defaults, and settings-label layers; after the fix the native error no longer reproduces.

Automated-run status: off-safety + crash-safety remain recorded green (2026-07-25 emulator run); the two positive engagement specs progressed through full remote-device sessions but a recorded end-to-end green is still pending on infra (remote-adb tunnel resets; MIUI install restriction on the backup phone; HF download throttling on the emulator after repeated fixture pulls). The session-layer plumbing they needed (`E2E_SYSTEM_PORT`, `E2E_ADB_SLOW`, `E2E_REMOTE_ADB_HOST`, sheet-close fallbacks, realistic budgets) is all landed, so the next run on a reachable device is just `--devices <id>`.

a-ghorbani:
## In-repo GGUF header reader verified on-device (Hermes)

`53b435a2` replaced `@huggingface/gguf` with a purpose-built header reader and deleted its accommodation stack (TextDecoder polyfill, metro browser-build redirect, exact pin, ESM jest mock). Because this feature's original failure mode was "green in Jest, dead on Hermes", the swap was re-proven at every layer:

- **Bundle**: releaseE2e APK builds with the metro redirect removed.
- **Hermes runtime**: the details sheet for the MTP fixture repo renders the **Speculative (MTP)** badge on the Android emulator — the probe's `DataView`/BigInt reads and hand-rolled UTF-8 work on-device with no `TextDecoder` global (screenshot below; Appium-driven flow, spec green).
- **Live remote validation**: the reader was run from Node against the four device-verified fixture files — Qwen3.5-MTP → `nextn=1`, gemma4-assistant draft → `nextn=4`, both non-MTP negatives clean.
- **Unit layer**: probe tests now drive byte-accurate handcrafted GGUF v1/v2/v3 headers (tokenizer-scale vocab arrays included) through a Range-honoring fetch mock; 267 suites / 4264 tests green.



Round-4/5 suggestions S1–S3 (library fetch cap, polyfill stream mode, decode perf) are obsoleted by the removal rather than fixed.

a-ghorbani:
## Visual evidence — speculative settings layout fix (38d1ddc4, Android emulator)

Two layout defects reported on the speculative section: dropdown items truncated by a fixed 170px width ("None (embedde…"), and long value labels squeezing the flex title/description column on the Draft Model and Draft Key/Value Cache Type rows.

After: the description columns wrap comfortably, and a long selected value ellipsizes inside its button (55% cap on the value side) instead of crushing the text column.



The dropdown now sizes to its longest label — both entries render in full:



a-ghorbani:
## Visual evidence — speculative settings redesign (bd1c4a9d, supersedes the 38d1ddc4 set)

Owner review rejected the width-juggling fix: the menu could overflow the screen edge and the descriptions stayed cramped. This pass changes the design instead: the Draft Model picker moves to its own full row under the title/description, and the two long speculative descriptions are cut to one clause each.

Draft Model as a full-row control — one-line description, complete picker label:



The dropdown anchors at the row's left edge, fully on-screen, both items complete:



Draft Key/Value Cache Type rows with the shortened copy:



a-ghorbani:
**Automated speculative e2e: recorded green runs (iOS simulator, branch builds)**

- `speculative.spec.ts` — **3/3 passing** (11m57s): off-safety (non-MTP model + speculative on), crash-safety (width-mismatched paired draft), and engagement (MTP model produces `draft_tokens > 0`).
- `speculative-paired.spec.ts` — **1/1 passing** (21m42s): separate 98 MB MTP draft paired with the 3.5 GB gemma-4-E2B target, speculative engagement confirmed.
- `quick-smoke` — passing on both Android emulator (47s) and iOS simulator (81s), covering the shared e2e helpers this PR touches on a non-speculative path.

Environment notes: the paired fixture (3.6 GB) exceeds the Android emulator's free disk, and the emulator's network throughput is too unstable for the model downloads, so the speculative runs are recorded on the iOS simulator. Real-device runs remain the follow-up for performance-level claims (GPU flash-attn path, acceptance rates).

Test fixes landed while getting these green: `bc70c80b` (Android keyboard dismissal emitted a system back that popped the screen behind the open search sheet — affected every Android spec driving HF search), `f958821f` (paired-spec download/test budgets are env-overridable for slow lanes).

a-ghorbani:
**Update: embedded speculative e2e also green on the Android emulator — 3/3.**

`speculative.spec.ts` on the Android emulator (fresh install, full download): off-safety ✓, crash-safety ✓, engagement ✓ with `draft: 63/159 (40%)` at 48.95 tokens/sec — speculative engagement is now proven on both platforms' native paths (Android CPU here, iOS sim earlier today).

The earlier emulator failures were environmental (a VPN was throttling the emulator's downloads to ~45 KB/s) plus one spec bug fixed in `d3b15aa1`: the footer Texts carry no accessibilityLabel, so Android exposes their content only as rendered text — and the uiautomator2 driver reports the missing attribute as the literal string "null", which defeated a plain truthiness fallback. The read now sanitizes and falls back to `getText()`.

Remaining device-only work is performance-level evidence (GPU flash-attn path, real acceptance rates on hardware).

PR #808, opened 2026-06-25
Add internet search to chat (web_search + read_url talents, BYOK)

a-ghorbani:
## Summary

Adds model-driven internet search to chat. Two new talents — `web_search` and `read_url` — are exposed to the model through the existing agent loop, so the model decides when to search the web or fetch a page during a turn. This is **not** a global pre-search: search is opt-in per Pal via `pact.talents`, and only Pals that declare the talents can use them.

### Provider-agnostic BYOK
- Bring-your-own-key design across multiple providers: **Tavily** (default), plus **Brave**, **Exa**, and **Parallel** behind a provider gate.
- Keys are stored in the device Keychain and read on demand — never bundled in the app and never proxied through a server.
- A small per-provider HTTP adapter layer normalizes each provider's wire format into a shared result shape.

### On-device budgeting
Raw provider hits are bounded on-device before reaching the model:
- result-count cap,
- snippet truncation on word boundaries,
- a token ceiling that drops trailing hits/page tail rather than cutting mid-fact,
- an in-session cache to avoid repeat fetches within a conversation.

### Settings + consent
- New **Internet Search** card in Settings for selecting the active provider and entering per-provider keys.
- First-enable consent gate before any network search can run.

### Docs
The relevant architecture docs (pals/talents, settings, agent-runner flows) were updated in the same change set to keep the architecture library in sync — no drift.

## Verification
- Unit tests cover the budgeting util, each provider adapter (including malformed/empty wire bodies), the talent engines, the registry wiring, and the provider store. Coverage meets the project threshold.
- Lint and typecheck pass.
- Visual evidence captured for the new Settings card and key-entry sheet.


a-ghorbani:
## Visual evidence — internet search (settings, consent, provider picker, key sheet, chat result chip + sheet)









a-ghorbani:
## Visual evidence — internet-search consent button placement

**Fix (`58397819`).** The first-enable consent gate reused the one-line settings-row layout (text left, button right, vertically centered). With its multi-line disclosure paragraph the "I understand" button floated mid-height beside a narrow wrapped text column — awkward on a phone.

Both disclosure blocks (the **"I understand"** gate and the consent-given **"Revoke"** state) now use a column layout: title → full-width paragraph → button below it, end-aligned. End-alignment uses `alignSelf: 'flex-end'`, so it flips correctly under RTL.

**Before:** button vertically centered beside the wrapped paragraph.
**After (shown):** button sits below the full-width paragraph, end-aligned. The consent-given/Revoke block received the identical treatment.



Captured on iOS simulator (Release e2e build) via the settings visual-capture spec.

a-ghorbani:
## Visual evidence — result-count copy and truncation fixes

Addendum for the latest UI-affecting commits (`9d51a6fc`, `44999867`):

- **Singular count**: a single-hit search now reads **"1 result"** instead of "1 results".
- **Truncation fix**: the chip label is split so a long query truncates while the count stays fully visible (previously a long query swallowed the count, e.g. "1 re…").



Captured on the iOS simulator (Release e2e build) from a seeded single-result web_search session.

Not visually capturable: the new link-open failure snackbar fires only when native `Linking.openURL` rejects, which cannot be deterministically induced on the simulator without mocking the native module; it is covered by the component's error-path handling instead. The remaining changes in these commits (read_url allowlist, hydration validation, accessibility labels) have no pixel impact.

PR #818, opened 2026-07-09
[Feat]: Discover llama.cpp /props for remote vision + context banners

a-ghorbani:
## Summary

Remote (OpenAI-compatible) models can be pointed at a self-hosted **llama.cpp** server. Two features were already coded but gated to on-device models only, and remote token counts were under-reported. This wires a `GET {baseUrl}/props` capability probe on llama.cpp servers and uses the response to unlock them:

- **Remote vision** — when `/props` reports `modalities.vision === true`, the chat image-attach affordance is enabled and attached images are inlined as `data:` URIs on the wire (a remote server cannot read the device filesystem). A non-vision server keeps attach disabled.
- **Remote context banners** — the server's real context window (`default_generation_settings.n_ctx`, top-level `n_ctx` fallback) drives the near-limit and context-full banners for remote models, replacing the weak hedged advisory. Length-truncation routes through the freshness-gated context-full branch, so a small `max_tokens` cap shows no false "context full".
- **Accurate remote token count (bug fix)** — remote completions now source `tokens_evaluated` / `tokens_predicted` from the llama.cpp `timings` object (`prompt_n` / `predicted_n`) already emitted on the finish chunk, so the context-usage ratio is correct instead of counting generated tokens only.

`/props` is llama.cpp-specific. It is fetched **only** when the persisted `serverType === 'llama.cpp'`; other server types (LM Studio / Ollama / vLLM / OpenAI) issue no `/props` request. The probe is best-effort — a timeout, non-2xx, or malformed body is a silent no-op that never breaks the models fetch or the connection. New `ServerConfig.contextLength` / `supportsVision` fields are optional and ride the already-persisted servers list, so old configs hydrate unchanged with no migration.

## Trust boundary

Remote vision is enabled purely from the server's `/props` self-report. This is the user's own configured server, and images already leave the device to it by the user's own attach action, so no new trust boundary is crossed. No separate remote-vision override ships here; the escape hatch is "don't attach an image," and a non-vision server's 400 surfaces on the existing completion-error path.

## Video-pal (real-time camera) note — considered and safe

`isMultimodalEnabled()` is the single read-point for both the one-shot chat image-attach path and the real-time video-pal camera path. Only the chat image-attach path is affected here: the video-pal `handleStartCamera` returns early on `!modelStore.context`, which a remote model always is, so the remote branch of `isMultimodalEnabled()` never reaches camera start. Only one-shot chat image-attach gains remote vision.

## Testing

- JS/TS-only change; no native surface touched (no `package.json` / `ios/` / `android/` / Podfile / gradle). No native build gate applies.
- 380 unit tests green across the touched suites (openai, ServerStore, ModelStore, bannerVariantResolver, BannerRow); typecheck + lint clean; coverage above threshold on every touched file.
- `/props` key names and `timings` token keys were pinned against a live llama.cpp build (vision `modalities.vision`, context `default_generation_settings.n_ctx`, tokens `prompt_n`/`predicted_n`), with an end-to-end proof: a real vision server answered "Red." about an attached `data:`-URI image. Visual evidence attached in a follow-up comment.

## Architecture docs

The cumulative architecture library (`context/architecture/remote-servers.md`, `chat-flow.md`, `model-loading.md`) lives in the team's control-plane repo, not in this app repo, so the same-PR doc update for this behaviour change was committed there alongside this branch rather than in this diff. The deltas cover: the two new `ServerConfig` capability fields + the `/props` discovery section and single-writer rule (remote-servers), the relaxed remote banner invariant + precedence and the `timings` token source (chat-flow), and the `isMultimodalEnabled` remote fallback with the `data:`-URI wire encoding (model-loading).

### DIFF
--- a/e2e/specs/features/remote-vision.spec.ts
+++ b/e2e/specs/features/remote-vision.spec.ts
@@ -1,0 +1,224 @@
+/**
+ * Remote Vision Capability Feature Tests
+ *
+ * Exercises the llama.cpp GET /props capability-discovery path end to end:
+ * a remote server that reports `modalities.vision` enables the chat image
+ * attachment affordance; a server that does not keeps it disabled. This is the
+ * user-visible surface of the /props -> ServerConfig.supportsVision ->
+ * ModelStore.isMultimodalEnabled -> ChatInput isVisionEnabled chain.
+ *
+ * The attach button is always rendered (ChatScreen showImageUpload={true}); only
+ * its enabled state flips with the discovered capability, so isEnabled() on the
+ * "Add image" button is the assertion. (Sending an actual image needs the native
+ * gallery picker, which is not automatable here; that path is covered by the
+ * committed visual capture.)
+ *
+ * Prerequisites:
+ *   - REMOTE_VISION_URL points at a llama.cpp server whose GET /props reports
+ *     modalities.vision === true (e.g. a SmolVLM build).
+ *   - REMOTE_NONVISION_URL (optional) points at a text-only server (no vision in
+ *     /props). When unset, the negative test is skipped.
+ *
+ * Environment variables:
+ *   REMOTE_VISION_URL         - vision-capable server (default http://192.168.0.62:1234)
+ *   REMOTE_VISION_MODEL_HINT  - partial model name in the picker (default 'SmolVLM')
+ *   REMOTE_NONVISION_URL      - text-only server (optional; negative test)
+ *   REMOTE_NONVISION_MODEL_HINT - partial model name for the text server (optional)
+ */
+
+import * as fs from 'fs';
+import * as path from 'path';
+import {expect} from '@wdio/globals';
+import {ChatPage} from '../../pages/ChatPage';
+import {DrawerPage} from '../../pages/DrawerPage';
+import {ModelsPage} from '../../pages/ModelsPage';
+import {Selectors, byPartialText} from '../../helpers/selectors';
+import {Gestures} from '../../helpers/gestures';
+import {TIMEOUTS} from '../../fixtures/models';
+import {SCREENSHOT_DIR} from '../../wdio.shared.conf';
+
+declare const driver: WebdriverIO.Browser;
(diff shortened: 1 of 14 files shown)


a-ghorbani:
## Visual evidence — remote llama.cpp vision





a-ghorbani:
## Test results

Recorded for posterity. All checks below were run against the merged branch (main merged in, so the build carries `NSLocalNetworkUsageDescription`).

### Static analysis
- **TypeScript** (`tsc --noEmit`): pass
- **ESLint** (changed files): pass, 0 warnings
- **Unit tests** (touched suites — `openai`, `ServerStore`, `ModelStore`, `BannerRow`, `bannerVariantResolver`): pass

### E2E — built from source, iPhone 17 Pro simulator (iOS 26) + Android emulator (API 36)

| Spec | iOS | Android |
| --- | :---: | :---: |
| `quick-smoke` (core journey) | ✅ | ✅ |
| `context-banner` (context-limit banner) | ✅ | ✅ |
| `talent-tool-use` | ✅ | ✅ |
| `remote-server` (vs live llama.cpp) | ✅ | ✅ |
| `remote-reasoning` (vs llama.cpp router) | ✅ | ✅ |
| `remote-vision` (**new** — see below) | ✅ | ✅ |

Remote specs were run against real `llama-server` instances: a SmolVLM build (`GET /props` → `modalities.vision: true`) and a multi-model router.

### New coverage added in this PR
`e2e/specs/features/remote-vision.spec.ts` — asserts the `GET /props` capability-discovery path end to end: a vision-capable remote server **enables** the chat image-attach button, and a text-only server keeps it **disabled** (`supportsVision → isMultimodalEnabled → ChatInput`). Env-configurable via `REMOTE_VISION_URL` / `REMOTE_NONVISION_URL`.

The image *send* itself is exercised by the committed visual capture (the native gallery picker is not automatable in Appium); the spec asserts the deterministic capability signal (attach-button enabled/disabled).

---

PR #819, opened 2026-07-10
Implement Phase 1 features including auth, sync, and backup systems

ghshhf:
## Description

Fixes # (issue number)

## Platform Affected

- [ ] iOS
- [ ] Android

## Checklist

  - [ ] iOS Simulator/Device
  - [ ] Android Emulator/Device
- [ ] Unit tests and integration tests pass locally.


a-ghorbani:
Thanks for the effort that went into this. I have to decline this one, though, mainly on product direction. PocketPal's core promise is that everything stays on the device: no account, no server, nothing collected.   Cloud sync of chats to Firestore inverts that. It requires sign-in, moves conversations off the phone, and makes the project the operator of a datastore holding users' chat history with the retention, deletion, and disclosure obligations that follow. That's a trade I don't want to make, even with client-side encryption in front of it.

PR #826, opened 2026-07-22
[Feat]: Add Polish and European Portuguese

a-ghorbani:
Wires Polish (`pl`) and European Portuguese (`pt`) into the language registry. Both locale files already landed via Weblate; this makes them selectable.

| locale | key coverage | translated |
| --- | --- | --- |
| `pl` | 97.0% | 93.1% |
| `pt` | 94.0% | 90.9% |

Untranslated keys fall back to English through `_.merge(enData, langData)`, so partial coverage degrades gracefully.

## Polish needs the Inter fallback

`pl` is added to `NON_LATIN_LOCALES` even though Polish is Latin script. The bundled Fraunces TTFs are the Latin-only Fontsource subset (199 codepoints) and stop at Latin-1 — they carry no Latin Extended-A, so `ą ć ę ł ń ś ź ż` and their capitals are missing: 16 of Polish's 18 diacritics. Verified by parsing the `cmap` table of `src/assets/fonts/Fraunces-{Regular,Medium}.ttf`. Inter covers all of them.

Without this, every Polish headline would render with missing glyphs. Portuguese needs only Latin-1 and stays on Fraunces.

The membership test for that list is **glyph coverage, not script family**. `yarn verify:fonts` now enforces it: for every wired locale not exempt, each letter used in its JSON must exist in the bundled Fraunces subset. Only letters are checked — the shared `— • … → ✓` punctuation is absent from Fraunces for every locale including `en` and is not a regression. A negative test in `scripts/__tests__/verify-fonts.test.js` removes `pl` from the exempt list and asserts the check fails.

## pt vs pt_BR

These are genuinely distinct locales, not duplicates: `ficheiros`/`arquivos`, `detetado`/`detectado`, `pormenores`/`detalhes`, enclitic `guia-o`/`te guia`; 25 `settings.*` keys differ. They happen to share both Settings header strings, so the e2e language spec cannot tell them apart — a unit test asserts they resolve to different translations instead of adding a scroll-dependent (flaky) e2e assertion.

## Verification

- `yarn test` — 262 suites, 3943 passed, 0 failed
- `yarn lint` — 0 errors; `yarn typecheck` — clean
- `yarn verify:fonts` — passes, and fails as expected when `pl` is removed from the exempt list
- `yarn l10n:validate` — all locale files valid
- Not a native change: no `package.json`, native module, `ios/`, `android/`, Podfile or build.gradle edits. Fonts already ship.

Visual evidence in a follow-up comment.

### DIFF
--- a/e2e/specs/features/language.spec.ts
+++ b/e2e/specs/features/language.spec.ts
@@ -57,6 +57,16 @@ const LANGUAGE_ASSERTIONS: Record<
     screenTitle: 'Tetapan',
     firstCardTitle: 'Tetapan Permulaan Model',
   },
+  pl: {
+    screenTitle: 'Ustawienia',
+    firstCardTitle: 'Ustawienia Inicjalizacji Modelu',
+  },
+  // pt shares both of these strings with pt_BR, so this spec proves the switch
+  // works but not that the two locales are distinct — locales.test.ts covers that.
+  pt: {
+    screenTitle: 'Configurações',
+    firstCardTitle: 'Configurações de Inicialização do Modelo',
+  },
   pt_BR: {
     screenTitle: 'Configurações',
     firstCardTitle: 'Configurações de Inicialização do Modelo',
@@ -80,7 +90,7 @@ const LANGUAGE_ASSERTIONS: Record<
 };
 
 // Order: start with non-English, end with English to restore default state
-const LANGUAGE_ORDER = ['fa', 'he', 'id', 'ja', 'ko', 'ms', 'pt_BR', 'ru', 'uk', 'zh', 'zh_Hant', 'en'];
+const LANGUAGE_ORDER = ['fa', 'he', 'id', 'ja', 'ko', 'ms', 'pl', 'pt', 'pt_BR', 'ru', 'uk', 'zh', 'zh_Hant', 'en'];
 
 describe('Language Switching', () => {
   let chatPage: ChatPage;
--- a/scripts/__tests__/verify-fonts.test.js
+++ b/scripts/__tests__/verify-fonts.test.js
@@ -177,6 +177,23 @@ describe('verify-fonts.js', () => {
     expect(result.output).toContain("add 'pl' to NON_LATIN_LOCALES");
   });
 
+  it('refuses to declare success when NON_LATIN_LOCALES cannot be parsed', () => {
+    // The glyph check is regex-driven, so the parse-failure branch is the
+    // safety net for the whole design: it must exit non-zero rather than
+    // quietly skipping coverage.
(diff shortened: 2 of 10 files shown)


a-ghorbani:
## Visual evidence — Polish + European Portuguese







a-ghorbani:
## Review round 1 — addressed

Independent review found one blocker and nine nits. Blocker plus the four highest-value nits are fixed in `f3f617e9`; the rest are documented in place.

**Blocking — architecture-doc drift (fixed, control-plane commit `c67a5b6`).** The plan pointed at `context/architecture/localization.md`, which does not exist, so the doc step silently no-op'd. The real docs are `theming.md` and `onboarding.md`, both of which hardcoded the fallback set and the now-wrong "primary script" criterion. Updated both, plus the `verify-fonts.js` description which claimed the script "only verifies presence" — it now enforces two guarantees.

**Fixed in code:**
- `scripts/sync-weblate.js` still listed 11 languages — the one enumeration site the original change missed. Download path would have silently skipped `pl`/`pt`.
- `typography.test.ts` never asserted the load-bearing positive case: that `pt`/`pt_BR` *stay* on Fraunces. Only the `pl`-falls-back direction was covered.
- `useTheme.test.tsx` had a second hardcoded fallback list, now derived from `NON_LATIN_LOCALES` so it can't rot again.
- The glyph check read only `Fraunces-Regular.ttf`; it now intersects coverage across all four cuts.
- Added the missing test for the parse-failure branch — the safety net for the whole regex-driven design, and previously the only untested path.
- The "how to add a language" comment listed 4 of 6 steps, omitting the dayjs map and the glyph check itself. That comment is the cheapest defence against the next locale repeating this bug.

**Documented rather than changed**, with reviewer agreement:
- Format-4 coverage doesn't resolve `idDelta`/`idRangeOffset`, so a codepoint mapped to `.notdef` would count as covered. This is the only direction the check can fail *unsafe*; verified zero divergence against a glyph-resolving parser across all 13 bundled TTFs, and now commented so a font re-cut doesn't quietly slip past.
- The format-12 `0x2fff` scan cap and the whole-file letter scan are both deliberate over-approximations that can only produce a spurious failure, never a false pass.

**Deliberately not fixed here:** the `Português (PT_B…` truncation. It's pre-existing (`pt_BR` truncated before this change) and is a fixed `width: 170` on `styles.menu`, not a naming problem. Fixing it is a Settings-layout change with its own visual-evidence obligation and regression risk; it doesn't belong in an l10n wiring PR.

Full suite after fixes: **262 suites, 3946 passed, 0 failed** (up 3 from the new assertions); lint 0 errors; typecheck clean; `verify:fonts` passes and still fails correctly when `pl` is removed from the exempt list.

PR #829, opened 2026-07-22
[Feat]: Searchable language picker in Settings

a-ghorbani:
Replaces the Settings language dropdown with the app's existing searchable bottom sheet.

## Problem

The picker did not scale with the locale count. Two defects, both visible in the PR #826 evidence:

- **Truncation** — a hardcoded `width: 170` on every menu item clipped labels: `Português (PT_B…`, `繁體中文 (ZH_HA…`. Wiring `pt` in #826 made it worse, putting the truncated `Português (PT_BR)` directly under `Português (PT)`.
- **Unmanaged height** — the menu had no `maxHeight`. At 14 locales it spanned nearly the full screen, with the last row against the screen edge. `de`, `et`, `fr` and `it` are already sitting unwired.

## Approach

No new component and no new dependency: `SearchableSelectSheet` already exists and has been in production since the TTS Supertonic picker needed to handle **31** languages. Same problem, larger list, already solved. Full-bleed rows make truncation impossible; a fixed snap point over a virtualised list bounds the height regardless of locale count.

Alternatives considered and rejected: capping the Paper `Menu` — hand-rolls content sizing, `maxHeight` and RTL anchor maths onto a primitive the design system is migrating away from, and its `width: 170` is shared by three other menus; the DS `Dropdown` — wraps the same Paper `Menu`, so it inherits the exact defect being fixed.

The design system has no language-picker frame, so this is not a parity slice — but the DS idiom for a multi-option setting is a row + value + chevron opening a scrimmed bottom card, which is what this is.

## Also fixed

Two real defects in the shared sheet, which also fix the TTS picker:

1. `handleSelect` called the `onClose` prop directly, bypassing the query reset — so reopening after a filtered selection showed a stale filtered list.
2. No `ListEmptyComponent` — a non-matching query rendered a blank sheet, with an `emptyText` style defined and unused.

Both change behaviour on a shipped surface, so both are covered by new unit tests and a manual smoke check.

## RTL

Text alignment is pinned to the layout direction, **not** `textAlign: 'auto'` — but the two element types need different expressions, which a forced-RTL device capture settled:

- **`Text`** (row label, empty state): plain `textAlign: 'left'`. React Native already mirrors `left`/`right` for `Text` under RTL, so `'left'` lands at the layout start in both directions. An `isRTL` ternary here **double-flips** and pushes rows to the layout end.
- **`TextInput`** (search field): `isRTL ? 'right' : 'left'`. `TextInput` gets no such mirroring, so the ternary is required. `auto` resolves to *natural* alignment from the first strong character, so in an RTL UI the field starts right-aligned under the Hebrew placeholder and then flips to the left the moment a Latin code is typed — which is the one interaction this design is built around. In an LTR UI the same setting would detach the `עברית (HE)` and `فارسی (FA)` rows to the opposite edge from every other row.

## Notes

- e2e testIDs are unchanged: `optionTestIDPrefix="language-option"` reproduces `language-option-<lang>` exactly, so `language.spec.ts` is byte-identical. Only the page object changed — it now waits on the sheet instead of a fixed `pause(500)`, and types the locale code to defeat list virtualisation.
- A new unit test asserts every display name carries its parenthesised locale code. That invariant is what makes search-by-code work, and it was previously unguarded — the existing test hardcoded 13 literals and omitted `fa`.
- `styles.menu` is deliberately untouched: it is shared by the key-cache, value-cache and search-provider menus, which keep the same truncation as a deferred cleanup.

## Verification

- `yarn test` — 262 suites, 3949 passed (baseline 3946), 0 failed
- `yarn lint` 0 errors · `yarn typecheck` clean · `yarn l10n:validate` · `yarn verify:fonts` pass
- Not a native change: no `package.json`, native module, `ios/`, `android/`, Podfile or build.gradle edit.

Visual evidence to follow in a comment.

### DIFF
--- a/__mocks__/external/@gorhom/bottom-sheet.js
+++ b/__mocks__/external/@gorhom/bottom-sheet.js
@@ -23,7 +23,18 @@ const BottomSheetBase = React.forwardRef((props, ref) => {
 // Mock FlatList that actually renders items
 const BottomSheetFlatList = React.forwardRef((props, ref) => {
   React.useImperativeHandle(ref, createMockRef, []);
-  const {data, renderItem, keyExtractor, ...restProps} = props;
+  const {data, renderItem, keyExtractor, ListEmptyComponent, ...restProps} =
+    props;
+
+  if (!data?.length && ListEmptyComponent) {
+    return React.createElement(
+      View,
+      {...restProps},
+      React.isValidElement(ListEmptyComponent)
+        ? ListEmptyComponent
+        : React.createElement(ListEmptyComponent),
+    );
+  }
 
   return React.createElement(
     View,
--- a/e2e/helpers/selectors.ts
+++ b/e2e/helpers/selectors.ts
@@ -384,6 +384,12 @@ export const Selectors = {
     get languageSelectorButton(): string {
       return byTestId('language-selector-button');
     },
+    get languageSheet(): string {
+      return byTestId('language-sheet');
+    },
+    get languageSearch(): string {
+      return byTestId('language-search');
+    },
     get displayMemoryUsageSwitch(): string {
       return byTestId('display-memory-usage-switch');
     },
--- a/e2e/pages/SettingsPage.ts
+++ b/e2e/pages/SettingsPage.ts
@@ -78,20 +78,24 @@ export class SettingsPage extends BasePage {
   }
 
   /**
-   * Tap the language selector button to open the language menu.
+   * Tap the language selector button and wait for the picker sheet to settle.
    */
   async openLanguageMenu(): Promise<void> {
     await this.tap(Selectors.settings.languageSelectorButton);
-    // Brief pause for menu animation
-    await browser.pause(500);
+    await this.waitForElement(Selectors.settings.languageSheet);
   }
 
   /**
-   * Select a language from the open language menu.
+   * Select a language from the open language picker.
+   * The list is virtualized, so the row is filtered into view by typing the
+   * language code before tapping it.
    * @param lang - Language code (e.g., 'en', 'id', 'ja', 'zh')
    */
   async selectLanguage(lang: string): Promise<void> {
+    await this.typeText(Selectors.settings.languageSearch, lang);
     await this.tap(Selectors.settings.languageOption(lang));
+    // The sheet animates out; a lingering backdrop would eat the next gesture.
+    await this.waitForElementToDisappear(Selectors.settings.languageSheet);
     // Wait for re-render after language change
     await browser.pause(1000);
   }
--- a/src/assets/icons/check-md.svg
+++ b/src/assets/icons/check-md.svg
@@ -1,3 +1,3 @@
-<svg preserveAspectRatio="none" width="100%" height="100%" overflow="visible" style="display: block;" viewBox="0 0 11.8658 8.53311" fill="none" xmlns="http://www.w3.org/2000/svg">
+<svg width="100%" height="100%" overflow="visible" style="display: block;" viewBox="0 0 11.8658 8.53311" fill="none" xmlns="http://www.w3.org/2000/svg">
 <path id="Union" d="M10.8418 0.17569C11.0761 -0.0585329 11.4561 -0.0585939 11.6904 0.17569C11.9243 0.409948 11.9243 0.789132 11.6904 1.02335L4.35738 8.35733C4.24492 8.46978 4.0916 8.53304 3.93257 8.53311C3.77373 8.53303 3.62115 8.46955 3.50874 8.35733L0.175736 5.02335C-0.0585785 4.78903 -0.0585787 4.41 0.175736 4.17569C0.410058 3.94147 0.789108 3.94141 1.02339 4.17569L3.93257 7.08389L10.8418 0.17569Z" fill="#181715"/>
 </svg>
--- a/src/components/LanguageSelector/LanguageSelector.tsx
+++ b/src/components/LanguageSelector/LanguageSelector.tsx
@@ -1,0 +1,66 @@
+import React, {useContext, useMemo, useState} from 'react';
+import {Text} from 'react-native';
+import {observer} from 'mobx-react';
+
+import {ChevronDownIcon} from '../../assets/icons';
+import {useTheme} from '../../hooks';
+import {AvailableLanguage, languageDisplayNames} from '../../locales';
+import {uiStore} from '../../store';
+import {L10nContext} from '../../utils';
+import {SearchableSelectSheet} from '../SearchableSelectSheet';
+import {Pressable} from '../ui/primitives/Pressable';
+
+import {createStyles} from './styles';
+
+/**
+ * App-language control: a content-sized trigger plus the shared searchable
+ * sheet. Search keeps every locale reachable by typing its Latin code even
+ * when the UI is in a script the user cannot read.
+ */
+export const LanguageSelector: React.FC = observer(() => {
+  const theme = useTheme();
+  const l10n = useContext(L10nContext);
+  const styles = createStyles(theme);
+  const [sheetOpen, setSheetOpen] = useState(false);
+
+  const options = useMemo(
+    () =>
+      uiStore.supportedLanguages.map(lang => ({
+        value: lang,
+        label: languageDisplayNames[lang],
+      })),
+    [],
+  );
+
+  return (
+    <>
+      <Pressable
+        testID="language-selector-button"
+        accessibilityRole="button"
+        accessibilityLabel={languageDisplayNames[uiStore.language]}
+        onPress={() => setSheetOpen(true)}
+        style={styles.trigger}>
+        <Text style={styles.triggerLabel} numberOfLines={1}>
+          {languageDisplayNames[uiStore.language]}
+        </Text>
+        <ChevronDownIcon
+          width={18}
+          height={18}
+          stroke={theme.colors.onSurface}
+        />
+      </Pressable>
+      <SearchableSelectSheet
+        testID="language-sheet"
+        searchTestID="language-search"
+        optionTestIDPrefix="language-option"
+        isVisible={sheetOpen}
+        onClose={() => setSheetOpen(false)}
+        title={l10n.settings.languageSheetTitle}
+        searchPlaceholder={l10n.settings.languageSearchPlaceholder}
+        options={options}
+        value={uiStore.language}
+        onSelect={value => uiStore.setLanguage(value as AvailableLanguage)}
+      />
+    </>
+  );
+});
--- a/src/components/LanguageSelector/index.ts
+++ b/src/components/LanguageSelector/index.ts
@@ -1,0 +1,1 @@
+export {LanguageSelector} from './LanguageSelector';
--- a/src/components/LanguageSelector/styles.ts
+++ b/src/components/LanguageSelector/styles.ts
@@ -1,0 +1,25 @@
+import {StyleSheet} from 'react-native';
+
+import {Theme} from '../../utils/types';
+
+export const createStyles = (theme: Theme) =>
+  StyleSheet.create({
+    // Content-sized: no width, so a long endonym is never clipped.
+    trigger: {
+      flexDirection: 'row',
+      alignItems: 'center',
+      flexShrink: 1,
+      gap: 8,
+      minHeight: 44,
+      paddingHorizontal: 14,
+      borderRadius: 12,
+      borderWidth: 1,
+      borderColor: theme.colors.outlineVariant,
+      backgroundColor: theme.colors.surface,
+    },
+    triggerLabel: {
+      flexShrink: 1,
+      color: theme.colors.onSurface,
+      fontSize: 16,
+    },
+  });
--- a/src/components/SearchableSelectSheet/SearchableSelectSheet.tsx
+++ b/src/components/SearchableSelectSheet/SearchableSelectSheet.tsx
@@ -61,7 +61,6 @@ export const SearchableSelectSheet: React.FC<SearchableSelectSheetProps> = ({
     return options.filter(o => o.label.toLowerCase().includes(q));
   }, [options, query]);
 
-  // Reset the query on every close path so a reopen starts unfiltered.
   const handleClose = () => {
     setQuery('');
     onClose();
@@ -100,10 +99,8 @@ export const SearchableSelectSheet: React.FC<SearchableSelectSheetProps> = ({
       onClose={handleClose}
       title={title}
       snapPoints={['75%']}
-      // Keep the header on screen when the keyboard opens. The default
-      // ("interactive") translates the whole sheet up by the keyboard height,
-      // which pushes the title and search field under the status bar; the list
-      // scrolls on its own, so extending is what we want.
+      // Default "interactive" slides the whole sheet up by the keyboard
+      // height, putting the header under the status bar.
       keyboardBehavior="extend"
       enablePanDownToClose
       enableContentPanningGesture={false}>
--- a/src/components/SearchableSelectSheet/SearchableSelectSheet.tsx
+++ b/src/components/SearchableSelectSheet/SearchableSelectSheet.tsx
@@ -100,6 +100,11 @@ export const SearchableSelectSheet: React.FC<SearchableSelectSheetProps> = ({
       onClose={handleClose}
       title={title}
       snapPoints={['75%']}
+      // Keep the header on screen when the keyboard opens. The default
+      // ("interactive") translates the whole sheet up by the keyboard height,
+      // which pushes the title and search field under the status bar; the list
+      // scrolls on its own, so extending is what we want.
+      keyboardBehavior="extend"
       enablePanDownToClose
       enableContentPanningGesture={false}>
       {isVisible ? (
--- a/src/components/SearchableSelectSheet/SearchableSelectSheet.tsx
+++ b/src/components/SearchableSelectSheet/SearchableSelectSheet.tsx
@@ -1,9 +1,10 @@
-import React, {useMemo, useState} from 'react';
+import React, {useContext, useMemo, useState} from 'react';
 import {View, Text} from 'react-native';
 import {BottomSheetFlatList} from '@gorhom/bottom-sheet';
 
 import {CheckMdIcon, SearchIcon} from '../../assets/icons';
 import {useTheme} from '../../hooks';
+import {L10nContext} from '../../utils';
 import {Pressable} from '../ui/primitives/Pressable';
 import {Sheet} from '../Sheet';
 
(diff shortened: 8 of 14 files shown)


a-ghorbani:
## Visual evidence — searchable language picker









a-ghorbani:
## RTL finding — the capture gate earned its keep

The forced-RTL capture surfaced something worth recording, and it is **not** a regression from this PR.

**What the capture shows:** under genuine `forceRTL`, the sheet chrome mirrors correctly (title, close button, search icon, and the search field's Latin query stays at the layout start), but the option rows do not: the check mark mirrors to the left while the label stays left-aligned, so the two collide on the selected row.

**Why:** `grep -rn "forceRTL\|allowRTL" src/` returns **nothing**. The app never enables RTL at runtime — `UIStore.setLanguage` only swaps l10n strings. So `I18nManager.isRTL` is `false` in every shipping build regardless of the selected language, and `textAlign: I18nManager.isRTL ? 'right' : 'left'` always resolves to `'left'`.

**Consequences, stated plainly:**

1. **No user impact today.** Hebrew and Farsi users already get an LTR layout with RTL text throughout the app. This picker behaves exactly like every other screen. The collision is only reachable by forcing RTL via `RCTI18nUtil_*` defaults on a simulator — a state production cannot enter.
2. **The `textAlign` pin is currently inert**, but it is still the correct expression and should stay. It costs nothing and is right the day RTL is enabled. The alternative (`textAlign: 'auto'`) would be actively wrong then — natural alignment flips on the first strong character, so a Latin query typed into an RTL field jumps edges mid-keystroke.
3. **The row/check collision is a latent issue for the whole app**, not this component. Enabling RTL is a much larger piece of work than this PR and should not be smuggled into it.

I am deliberately not "fixing" this here: the only real fix is to enable RTL app-wide, which is out of scope and would make this diff unreviewable.

Captures 08 and 09 are included above as the record. The simulator has been reverted to LTR.

a-ghorbani:
## Keyboard-open state — sheet header clears the status bar (pre-existing, shared Sheet)



a-ghorbani:
## Keyboard-open state — real, pre-existing, and NOT fixed here

Raised from review of the evidence above: when the keyboard opens, the sheet is pushed up far enough that its header renders under the status bar (visible in the empty-state capture — the "Language" title collides with the clock).

**Confirmed real, and it is in the shared `Sheet`, not this picker.** `src/components/Sheet/Sheet.tsx` sets neither `topInset` nor `keyboardBehavior`, so gorhom's default interactive keyboard handling translates the sheet up by the keyboard height with nothing bounding it at the top. The existing `maxDynamicContentSize` guard does not cover this: it only applies when `enableDynamicSizing` is on, i.e. when no `snapPoints` were supplied — and `SearchableSelectSheet` supplies `['75%']`.

So this affects **every sheet in the app that contains a text input**, and predates this PR. It is newly *visible* here only because this PR puts a search field on a Settings surface.

**I attempted the obvious fix and backed it out.** Adding `topInset={insets.top}` to the shared `BottomSheetModal` looked correct and typechecked, and all 6 sheet/consumer unit suites stayed green — but on device the sheet then **failed to open at all**. Reproduced across three runs; reverting and rebuilding restored it (captures re-run, 3/3 passing). Percentage snap points combined with `topInset` appear to resolve to a geometry the sheet will not present.

The unit tests would have let that ship. Only the device run caught it.

I have deliberately left this unfixed rather than ship a half-understood change to a component every sheet depends on:

- The correct fix is probably `keyboardBehavior="extend"` (expand toward the top snap point instead of translating the whole sheet), possibly with an explicit top snap point — but that changes keyboard behaviour for every existing sheet and needs its own design pass and its own captures.
- Doing it here would make this diff unreviewable and put an unrelated regression risk on a locale PR.

Recommend a separate task against the shared `Sheet`, with device captures for each consumer that has an input. `Sheet.tsx` is untouched by this PR (`git diff origin/main..HEAD -- src/components/Sheet/Sheet.tsx` is empty).

a-ghorbani:
## Keyboard-open state — FIXED (keyboardBehavior=extend)




a-ghorbani:
## Keyboard-push: fixed — correcting my earlier comment

My previous comment said this was pre-existing in the shared `Sheet` and should be deferred. **Half right, and I gave up too early.** There is a per-consumer fix that does not touch the shared component.

`Sheet` spreads `{...props}` onto `BottomSheetModal` *after* its own defaults, so a single consumer can override keyboard behaviour. `SearchableSelectSheet` now passes `keyboardBehavior="extend"`: the sheet expands toward its snap point instead of translating up by the keyboard height, so the header and search field stay put and the keyboard simply overlays the bottom of the list. The list already scrolls on its own, so nothing becomes unreachable.

Verified on device — see the captures above: the title now clears the status bar with the keyboard open. `git diff origin/main..HEAD -- src/components/Sheet/Sheet.tsx` is **empty**; blast radius is this component and the TTS picker, both re-captured/tested.

### Why not `topInset` on the shared `Sheet`

For the record, since it looks like the obvious fix: adding `topInset={insets.top}` typechecked and kept all sheet unit suites green, but on device **the sheet then failed to open at all** (reproduced 3×, reverted, re-verified). Percentage snap points plus `topInset` resolve to a geometry the sheet will not present. Unit tests would have shipped it.

### Still deferred, and now enumerated

The underlying default still affects every other sheet with a text input. Sheets that pass explicit `snapPoints` (so the `maxDynamicContentSize` guard does not apply) **and** contain an input:

| Sheet | snapPoints |
| --- | --- |
| `RemoteModelSheet` | `80%` |
| `AuthSheet` | `85%` |
| `ModelErrorReportSheet` | `80%` |
| `ContentReportSheet` | `70%` |
| `ServerDetailsSheet` | `70%` |
| `HFTokenSheet` | `60%` |
| `SearchProviderKeySheet` | `50%` |

`EnhancedSearchBar` and `PalSheet` pass no snap points, so `enableDynamicSizing` + `maxDynamicContentSize` already bound them.

Note `HFTokenSheet` wraps its body in `Sheet.ScrollView` (`BottomSheetKeyboardAwareScrollView`, built on `react-native-keyboard-controller`) — the repo's existing keyboard idiom, which mitigates but does not re-anchor the header. That library ships no FlatList equivalent, which is why the virtualized select list needed the `keyboardBehavior` route instead.

Recommend a follow-up applying one consistent approach across those seven, with device captures each. Not this PR.

a-ghorbani:
## e2e verified on both platforms — 14/14 locales each

The language spec now runs green against the new picker on both platforms:

| Platform | Result | Duration |
| --- | --- | --- |
| iOS (sim) | ✅ 1 passing — 14/14 locales | 2m 17s |
| Android (emulator) | ✅ 1 passing — 14/14 locales | 1m 55s |

Every locale switches and both assertions hold (`fa`, `he`, `id`, `ja`, `ko`, `ms`, `pl`, `pt`, `pt_BR`, `ru`, `uk`, `zh`, `zh_Hant`, back to `en`).

**`e2e/specs/features/language.spec.ts` is byte-for-byte unchanged** — `git diff origin/main..HEAD` on it is empty. That was the design goal and it held, because `optionTestIDPrefix="language-option"` reproduces the original row testIDs exactly, including `language-option-pt_BR`.

Only the page object changed, and each change is forced by the new interaction model rather than cosmetic:

1. `openLanguageMenu()` — waits on the additive `language-sheet` testID instead of a fixed `browser.pause(500)`. The old Paper `Menu` rendered synchronously; a sheet animates in.
2. `selectLanguage()` — types the locale code into `language-search` before tapping. `BottomSheetFlatList` virtualizes, so a row that has not rendered is not tappable at all. This works because every display name embeds its uppercase code and the filter lowercases both sides — the invariant the new `locales.test.ts` guard now pins.
3. `selectLanguage()` — waits for the sheet to disappear before returning. The sheet has a dismiss animation the popup never had, and a lingering backdrop swallows the next iteration's scroll gesture. This is a flake the old implementation could not have, and it would have surfaced intermittently across 14 sequential switches rather than failing cleanly.

The POM's public API is unchanged, so no other spec is affected.

a-ghorbani:
## Dark mode — sheet, selected-row check, filtered




a-ghorbani:
## Dark mode captured — and it corrects a claim I made earlier

§10 requires light **and** dark captures; dark is now posted above. It also disproves something I asserted in this thread.

**What I claimed:** `check-md.svg` has a hardcoded `fill="#181715"` and the component passes `stroke`, which cannot recolour a fill-only path — so on the `#000000` dark background the selected-row check would be **invisible**.

**What the device shows:** the check is clearly visible in dark mode. I was wrong.

**Why:** `stroke={theme.colors.primary}` does have an effect — react-native-svg applies it to the path. In dark mode `primary` is `#DADDE6`, so the glyph is outlined in near-white against the dark fill and reads perfectly well. My reasoning was right about the hardcoded fill and wrong about the consequence, because I forgot the stroke still paints even when the fill does not change.

**What is actually true, and it is cosmetic:** the check renders as a *solid* mark in light mode (fill `#181715` ≈ `LIGHT_PRIMARY` `#333333`, so the discrepancy is invisible) and as an *outlined* mark in dark mode (fill stays `#181715`, stroke `#DADDE6`). Same glyph, slightly different weight between themes. Nobody would file a bug about it.

I am **not** changing the asset in this PR. It is pre-existing, affects the TTS picker identically, is not a defect in any user-visible sense, and making it properly themeable (`fill="currentColor"` + `color=` prop) is an icon-system change — only 5 of our 140 SVGs use `currentColor`, so doing it for one icon would be an inconsistency of its own.

Also visible in the second capture: `keyboardBehavior="extend"` holds in dark mode — header and search field stay clear of the status bar with the keyboard up.

a-ghorbani:
## TTS smoke check — cannot be performed, and that corrects a risk claim I made

I said the two `SearchableSelectSheet` fixes "change behaviour on a shipped 31-option picker" and committed to a manual smoke check of `tts-hero-language-picker`. **That premise was wrong.**

`TTSSetupSheet` is exported from the components barrel but **never mounted anywhere in the app**:

- `grep -rn "<TTSSetupSheet" src/` → only its own two test files
- Same on `origin/main` — this is not something this branch changed
- `HeroRow` is referenced only inside `src/components/TTSSetupSheet/`
- No sheet registry, no dynamic host, no screen mounts it (`SheetHost`/`registerSheet` grep is empty; the only host is `HubRunSheetHost`, unrelated)

So there is no reachable second consumer, and no UI path to smoke. The real blast radius of the shared-component changes is **the language picker only** — which is fully covered above by device captures and 14/14 e2e on both platforms.

What I can verify, and did:

```
PASS src/components/SearchableSelectSheet/__tests__/SearchableSelectSheet.test.tsx
PASS src/components/TTSSetupSheet/__tests__/TTSSetupSheet.integration.test.tsx
PASS src/components/TTSSetupSheet/__tests__/TTSSetupSheet.test.tsx
PASS src/components/TTSSetupSheet/__tests__/HeroRow.test.tsx
PASS src/components/TTSSetupSheet/__tests__/VoicePickerView.test.tsx
Test Suites: 5 passed · Tests: 32 passed
```

The TTS picker's unit and integration coverage exercises the changed component and stays green.

**For the reviewer:** `what.md` §4d and my earlier comment both describe the TTS picker as a shipping surface. Treat that as corrected — the risk it was guarding against does not exist, though the fixes remain correct on their own merits. Whether an unmounted `TTSSetupSheet` is intentional (staged for a future entry point) or dead code is outside this PR's scope, but worth someone's attention.

a-ghorbani:
## RTL CORRECTED — rows now align to the layout start




a-ghorbani:
## Review round 1 — both blockers fixed. My RTL diagnosis was wrong.

### The RTL pin was inverted, and my earlier comment misdiagnosed it

Review caught that `textAlign: I18nManager.isRTL ? 'right' : 'left'` is **wrong for `Text`**. React Native already mirrors `left`/`right` for `Text` under RTL — confirmed in `RCTTextAttributes.mm`:

```objc
if (_layoutDirection == UIUserInterfaceLayoutDirectionRightToLeft) {
  if (alignment == NSTextAlignmentRight) { alignment = NSTextAlignmentLeft; }
  else if (alignment == NSTextAlignmentLeft) { alignment = NSTextAlignmentRight; }
}
```

So the ternary **double-flipped**: it yielded `right`, RN flipped it to `left`, and rows landed at the layout *end* — the exact ragged edge the invariant exists to prevent.

**Fix:** plain `textAlign: 'left'` for `Text` (row label + empty state), which resolves to the layout start in both directions. The search field keeps the ternary — `TextInput` gets no such mirroring, which is why the same expression behaved differently for the two elements in one screenshot.

**Corrected RTL captures posted above.** Rows now align right (layout start), the check mark sits left (layout end), nothing collides. Re-verified LTR is unchanged, and the full suite is still green (262 suites / 3949 passed).

**Retracting my earlier explanation.** I previously wrote that the left-aligned rows were "because the app never enables RTL, so `isRTL` is always false." That was wrong and self-contradictory — the same capture showed mirrored chrome, which requires `isRTL === true` (I had forced it via `RCTI18nUtil` defaults). The rows were misaligned because of the double flip, not because RTL was inactive. It remains true that RTL is never enabled in production, so there was no user impact either way.

The architecture doc has been corrected to record the `Text` vs `TextInput` distinction, so the next person doing app-wide RTL does not repeat this.

### Public-artifact hygiene

Removed the internal doc anchors from the PR body and added the missing `enhancement` label.

### On the non-blocking points

- **Keyboard-occluded rows** (`extend` does not pad the list for keyboard height): real, and I am leaving it. Typing filters the list, which is the primary path, and the alternative — padding `listContent` by keyboard height — is a change to a shared component that deserves its own captures across both consumers. Better as a follow-up than bolted on here.
- **No before/after pairing:** fair. The "before" is PR #826's evidence; a side-by-side would have been more convincing.
- **`useMemo(…, [])`, trigger a11y polish, duplicate `noResults` source string, `check-md.svg` fill:** all agreed, all cosmetic, none touched here.

a-ghorbani:
## Check mark — no longer stretched



a-ghorbani:
## Check-mark icon was stretched — fixed

Reported in review of the captures: the tick looked oversized and distorted. It was.

`check-md.svg` is a Figma export carrying `preserveAspectRatio="none"` with a **1.39:1** viewBox. Rendering it at `width={20} height={20}` therefore *stretched* it to fill a square — 39% taller than its natural proportion, with correspondingly uneven stroke weight, rather than fitting inside the box. Removing that one attribute restores the default `xMidYMid meet` and the glyph fits correctly. Recaptured above.

Worth flagging beyond this PR: **75 of 140 icons in `src/assets/icons/` carry the same `preserveAspectRatio="none"` artifact**, and many have non-square viewBoxes — `chevron-down-*` (1.77), `chevron-left/right-*` (0.57), `alert-triangle-*` (1.14), the other `check-*` sizes (1.39). Any of those rendered at equal width/height are distorted the same way. I have fixed only the icon this PR renders; a sweep is its own task.

## Comment load reduced, and the recurrence fixed at the source

Also raised: over-commenting shows up on nearly every PR and has to be prompted for each time.

Applied to this diff — comments in `src/` went from **20 lines to 8**. The cuts, by cause:
- `handleClose`'s "reset the query on every close path" — the function is named `handleClose` and its body is `setQuery('')`. Stating the obvious.
- The empty-state `textAlign` note — echoed the rule already stated two properties above.
- A test-body comment restating what the `it(...)` name said.
- The two RTL notes and the `keyboardBehavior` note were real, but 4–6 lines each where 2–3 carried the whole fact.

What survives is the case that earns it: RN has no `textAlign: start`, and auto-mirrors `left`/`right` for `Text` but not `TextInput`. Without that note the asymmetry looks like sloppiness and gets "cleaned up" straight back into the bug this PR fixed.

The recurrence itself is now addressed in the control plane rather than per-PR: the guidance existed but never reached the thing writing the code, so it is now a first-class section in the agent instructions, a working rule for the implementer, and a required review lens where a finding must name *which* cause — because "too many comments" is not actionable, and "this comment exists because the function does three things" is.

a-ghorbani:
## Correction: the TTS picker IS reachable — my earlier comment was wrong

An earlier comment claimed `TTSSetupSheet` is "never mounted anywhere in the app" and concluded the shared-component changes had only one live consumer. **That is false. Disregard it.**

`<TTSSetupSheet />` is mounted at the app root — `App.tsx:216`, inside `BottomSheetModalProvider` — and is opened by `ttsStore.openSetupSheet()` from two user-facing controls:

- `src/components/VoiceChip/VoiceChip.tsx:67,78` — the voice chip in the chat input (`ChatInput.tsx:644`)
- `src/components/TextMessage/PlayButton.tsx:73` — the play button on an assistant message

So the path is real: **VoiceChip → TTSSetupSheet → VoicePickerView → HeroRow → SearchableSelectSheet** (the language picker appears when the Supertonic engine is selected).

**How I got it wrong:** I grepped `src/` for the mount point. `App.tsx` lives at the **repo root**, not under `src/`, so the search could not have found it. I then reported the absence of a result as a positive finding, and wrote it into a public comment. That is the same failure mode as the Hermes `main.jsbundle` grep earlier in this PR — treating "my search returned nothing" as "the thing does not exist" — and I did not run a control to check the search could find anything at all.

**Consequences to re-state correctly:**

1. The original risk framing in the design doc was right: the two `SearchableSelectSheet` fixes (query reset on the selection path, `ListEmptyComponent`) **do** change behaviour on a shipping, user-reachable surface. So do `keyboardBehavior="extend"` and the `textAlign` corrections.
2. The TTS smoke check is therefore **required**, and it is possible. It was never blocked; I just looked in the wrong place. Running it now — result to follow in a separate comment.

a-ghorbani:
## TTS setup sheet — second consumer's entry point works



a-ghorbani:
## TTS smoke check — done, with an honest limit

Ran it on device now that I know the surface is reachable. Path: chat input voice chip → `ttsStore.openSetupSheet()` → `TTSSetupSheet` (mounted at `App.tsx:216`).

**Result:** the sheet opens and renders correctly — Kitten / Kokoro / Supertonic engine rows, all intact. Capture above. Nothing in the shared-component changes broke the second consumer's entry point.

**What I could not exercise, and why:** the Supertonic *language* picker — the part that actually instantiates `SearchableSelectSheet` — is gated on `supertonicDownloadState === 'ready'` (`HeroRow.tsx:87-89`). That needs a ~380 MB model download on the simulator. So the changed component is not reachable in this flow on a clean device without a real download.

**What does cover it:** `HeroRow.test.tsx` has a dedicated `describe('language picker')` block that drives the real code path with the store forced to `ready` — asserting the trigger appears only for Supertonic-and-ready, that tapping it opens `tts-language-sheet`, that `tts-language-option-na` ("Auto") is first, and that the localized language names render. Those pass, along with the rest of the TTS suites (5 suites / 32 tests).

So: entry point verified on device, the picker itself verified by tests that exercise the same component with the same props. I would rather state that split plainly than claim a full manual smoke I did not perform.

PR #830, opened 2026-07-22
[Bug]: scope the llama.cpp /props capability probe to a model

### LINKED ISSUE #828: [Bug]: /props capability probe omits the model param, so discovery fails on multi-model llama.cpp routers
## Summary

The llama.cpp capability probe requests `GET {baseUrl}/props` with no model parameter. A single-model `llama-server` answers with the loaded model's real properties, so this works. A multi-model router / proxy in front of llama.cpp (llama-swap style, one that lazily starts a server per model) answers bare `/props` with a placeholder describing nothing, so every discovered capability comes back empty.

Introduced with the `/props` discovery added in #818 — that PR was validated against a single-model server, where the behaviour is correct.

## Evidence

Against a multi-model router (`Server: llama.cpp`, ~45 models via `/v1/models`):

```
GET /props
  model_path : "none"
  modalities : null
  n_ctx      : 0

GET /props?model=gemma-4-e2b
  modalities : {"vision": true, "video": true, "audio": true}
  n_ctx      : 8192
```

The same model reports full vision support when the request is scoped to it.

## Impact

Server-type detection is not the problem — the router does send `Server: llama.cpp`, so the probe fires. It just receives a contentless response, and both consumers of the probe degrade silently:

- **Remote vision** — `supportsVision` never becomes true, so the chat image-attach affordance stays disabled for a vision-capable model.
- **Remote context banners** — `n_ctx` is `0` instead of the real window, so the near-limit and context-full banners lose the accurate context length that #818 introduced.

Both fail closed rather than misbehaving, and single-model servers are unaffected — which is why this went unnoticed.

## Suggested fix

Pass the selected model id to the probe, e.g. `GET {baseUrl}/props?model=<id>`, falling back to the bare request when no model is selected yet. Call site: `fetchServerProps(server.url, apiKey, PROPS_TIMEOUT_MS)` in `src/store/ServerStore.ts` (llama.cpp-only branch), implemented in `src/api/openai.ts`.

Ordering caveat: the probe currently runs once per server right after the models list is stored, while the capability is really per model. On a multi-model server, switching the selected model should re-probe, otherwise the cached `supportsVision` / `contextLength` describe whichever model was probed first.

## Reproduce

1. Point a remote server at a multi-model llama.cpp router that exposes several models via `/v1/models`.
2. Add it in the app and select a vision-capable model.
3. Expected: image-attach enabled and the real context window drives the banners. Actual: attach disabled, context window `0`.

`e2e/specs/features/remote-vision.spec.ts` reproduces this — it passes against a single-model vision server and fails against a router for this reason, not a product regression in the chat path.

a-ghorbani:
## Summary

Fixes #828.

The llama.cpp `GET /props` capability probe asked the server for its properties without naming a model. A single-model `llama-server` answers with the loaded model, so this worked. A multi-model router answers the bare form with a placeholder (`model_path: "none"`, `n_ctx: 0`, no `modalities`), so every discovered capability came back empty: image attach stayed disabled for a vision-capable model, and the context banners lost their real window.

Capabilities are per model, so they are now probed and stored per model, resolved against the backend the live session is actually bound to, and answered through a single selector shared by every consumer.

`/props` can only answer for a model the server has **loaded**, though — and on a swap router `?model=` *loads* it, so browsing 47 models cannot mean starting 47 of them. The `GET /v1/models` body the app already fetches turns out to state the same facts per row, so it now feeds the declared axis as a second, non-probing source: the model card and the add sheet can say what a remote model supports **before** anything is activated, with no extra request.

### Per-model capability discovery

- `src/api/openai.ts` — `fetchServerProps` takes an optional model id and scopes the request to `?model=<id>` (URL-encoded). It returns per-model caps and only ever sets a field from a body that describes an actually loaded model: `contextLength` only for a finite `n_ctx > 0`, `supportsVision` only when the body names a real model. A router placeholder yields nothing. The probe's default bound is the props timeout rather than the 30 s connection default.
- `src/store/ServerStore.ts` — new `remoteCaps` map keyed by the full model id (`<serverId>/<remoteModelId>`), persisted alongside `remoteReasoning`, with `fetchRemoteModelCaps` as its only writer. It merges field-wise, so a partial answer never blanks a known field, and a probe that resolves nothing writes nothing. At most two requests: the scoped probe, plus a bare retry only when the server is provably serving that one model — on a multi-model server the bare form describes whichever model is resident, so it is never issued there. `removeServer` prunes caps by server prefix.
- `src/store/ModelStore.ts` — the probe fires on remote-model activation, detached, instead of once per server after the models list is fetched. That also removes the race where a foreground refresh could clobber a good per-model result.

### Binding capabilities to the session's backend

- A capability entry records the url it was probed against. An entry that describes a different backend than the live session is bound to resolves to unknown rather than being trusted, which fails closed: editing a server's url or type drops the stale entries instead of reporting another server's answers, and the cached model list is dropped when a server is repointed.
- Unknown capabilities are re-probed when the app returns to the foreground.

### One resolver for every capability question

- `src/utils/modelCaps.ts` (new) — `resolveModelCaps(model, env)`, pure and synchronous, is now the only place `model.origin` is branched on to answer a capability question. It reports two independent axes: what the model declares (`vision`, `contextLength`), and what the session that exists right now can do (`visionActive`, `effectiveContextLength`). Session facts apply only to the active model on **both** origin legs, so a card can never borrow another model's load state. Both length fields are `> 0` or absent, so callers do not re-check.
- `src/store/ModelStore.ts` — assembles the resolver input in one private place and exposes `activeModelCaps` and `capsFor(model)`. No component builds that input or reads `remoteCaps` itself.
- `ModelStore.isMultimodalEnabled` is **deleted**. It was an async method that could repair `isMultimodalActive` from a reader, which made the flag a self-correcting cache that no writer had to keep complete. With it gone the flag has only its two transition writers and is a maintained observable; no reader performs a native re-verify. The attach affordance and the send-path image gate now read the same field of the same call, so they cannot disagree by ordering.
- `ChatScreen` drops the `useState`/`useEffect` pair that resolved local vision asynchronously; the value is read in the `observer` render body, so a capability landing from the detached probe or a model load re-enables attach with no further user action. `useChatSession`, `VideoPalScreen` and `BannerRow` read the same selector.

### The models list as a second, non-probing capability source

- `src/utils/listCaps.ts` (new) — `deriveListCaps(row, serverType)` is the only interpreter of a `/v1/models` row: pure, synchronous, and yielding at most `supportsVision` and `contextLength`. Vision from `architecture.input_modalities` containing `image` — the array always carries `text`, so a missing `image` is the server answering "no", not failing to answer — else, only if that key was absent, from `capabilities` containing `multimodal`. Context from `meta.n_ctx`, else the last `--ctx-size`/`-c` in `status.args` in either the space or `=` form, admitted only as an integer `> 0`. No other argument is read, and every failure lands on an absent field rather than a default (`--ctx-size 0` means "the model's trained window", which is unknown, not `0`). The `serverType === 'llama.cpp'` gate lives inside the function, because it has two callers with two different sources for that type. `status.args` can carry filesystem paths, so this is the only thing permitted to read it: nothing logs, persists or serialises a raw row.
- `src/api/openai.ts` — a single-model `llama-server` describes its model twice, and only the sibling `models[]` array carries `capabilities`; the fetch joins that entry onto its `data[]` row by `name ?? model`, so a direct server needs no extra request either. Servers emitting no `models[]` are unaffected.
- `src/store/ServerStore.ts` — `listCaps` is a **computed** with no writer and no persistence. `serverModels`'s lifecycle is already the right invalidation: replaced by each fetch, dropped when a server's url or type changes, dropped with the server. It therefore needs no probed-url check — a list cannot outlive the url it came from.
- Precedence is field-wise, probe first, and the list reaches the **declared** axis only. `visionActive` and `effectiveContextLength` are computed from the probe result before the list is consulted, so no derived value can gate attach, the send-path image gate, the video-pal camera start or a context banner. That holds structurally rather than by convention: there is no write path from the list into `remoteCaps`, and the two types carry a literal `tier` discriminant making them mutually non-assignable.

### Capabilities on the model card and the add sheet

- `src/screens/ModelsScreen/ModelCard/ModelCard.tsx` — remote cards gain the expand affordance they never had, so the details block is reachable for the first time. The expanded block is server-sourced only: the full model name, the discovered context window when known, and a vision indicator that states `Supported` / `Not supported` / `Unknown` explicitly — unknown is never expressed as absence. The header glyph is sourced from the same resolver as the cell, so the two cannot contradict each other. Two rows that would otherwise render device claims about someone else's hardware are suppressed for remote: the memory estimate (which reads "0 B" with no local file) and the Author cell (which holds the server name, already shown in the header chip). Local cards are unchanged in every state.
- `src/components/RemoteModelSheet/RemoteModelSheet.tsx` — against a llama.cpp server each row carries a **three-state** vision slot: the eye glyph, the text-chat glyph, or a muted em-dash for "this build does not say", with an accessibility label on all three. Without the third state "no vision" and "cannot tell" would collide, and a screen reader would hear nothing for either. Any other server type renders no slot at all. The slot reads the persisted type of the selected server, not the type the sheet detected, since pressing a known-server chip never sets that state.
- Card roots fall back to the model id when there is no filename, and the new cells and rows carry per-model testIDs, so cards and rows are individually addressable.
- Three new English strings; other locales are owned by Weblate.

## Verification

- Lint (0 errors), typecheck, and the full jest suite green: **265 suites, 4096 tests**, coverage 76.2% overall; the capability resolvers at 100% (`modelCaps`, `remoteCaps`) and 95% (`listCaps`).
- New unit coverage: scoped/encoded request URL, placeholder rejection, `n_ctx` zero/negative/non-numeric, vision decided only from a model-describing body, timeout fallback; store-side field-wise merge, sibling models not inheriting each other's caps, the single-model gate matrix, prefix prune, stale-binding invalidation; the resolver's tri-state, precedence and active-model-only rules; the list parser against verbatim captured router and direct-server bodies, including both argument forms, the short flag, last-occurrence-wins, and every documented parse failure; the join and its degenerate one-row case; card and add-sheet scenarios for pre-activation state, the three slot states being mutually distinct, the non-llama.cpp no-op, and cold-start fill-in with no user action.
- Mutation-checked: replacing the shared test store's capability resolution with a fixed answer fails 22 tests across the chat, banner and card suites; removing the type discriminant fails typecheck; reverting the header-glyph source or ungating the vision label each red their own cases. The suites are gates, not decoration.
- Live verification against a multi-model llama.cpp router (47 models) and a direct single-model `llama-server`, on **both** platforms: five models are marked as vision-capable in the list and on the cards before anything is activated, an expanded never-activated card states its context window and vision support, activating a vision model enables attach and its text-only sibling on the same server disables it, and a server of another type produces byte-identical screens to before.
- Visual evidence for the chat surface, the model card, and the pre-activation list-sourced surfaces is posted as PR comments.

### DIFF
--- a/__mocks__/stores/modelStore.ts
+++ b/__mocks__/stores/modelStore.ts
@@ -293,11 +293,6 @@ class MockModelStore {
     return this.availableModels.some(model => model.id === modelId);
   }
 
-  async isMultimodalEnabled(): Promise<boolean> {
-    // Mock implementation - return false by default for tests
-    return false;
-  }
-
   async getModelFullPath(model: Model): Promise<string> {
     // Mock implementation - return a simple path for tests
     return `/mock/path/${model.filename || model.name}`;
--- a/__mocks__/stores/modelStore.ts
+++ b/__mocks__/stores/modelStore.ts
@@ -3,11 +3,21 @@ import {computed, makeAutoObservable, observable} from 'mobx';
 import {modelsList} from '../../jest/fixtures/models';
 
 import {downloadManager} from '../services/downloads';
+import {mockServerStore} from './serverStore';
 
-import {Model, ContextInitParams} from '../../src/utils/types';
+import {
+  Model,
+  ContextInitParams,
+  RemoteSessionBinding,
+} from '../../src/utils/types';
 import {LlamaContext} from 'llama.rn';
 import {CompletionEngine} from '../../src/utils/completionTypes';
 import {createDefaultContextInitParams} from '../../src/utils/contextInitParamsVersions';
+import {resolveModelCaps} from '../../src/utils/modelCaps';
+import type {
+  CapabilityEnv,
+  ModelCapabilityView,
+} from '../../src/utils/modelCaps';
 
 class MockModelStore {
   models = modelsList;
@@ -20,6 +30,9 @@ class MockModelStore {
   isStreaming = false;
   context: LlamaContext | undefined = undefined;
   engine: CompletionEngine | undefined = undefined;
+  isMultimodalActive: boolean = false;
+  activeContextSettings: ContextInitParams | undefined = undefined;
+  activeRemoteBinding: RemoteSessionBinding | undefined = undefined;
 
   // Memory calibration variables
   availableMemoryCeiling: number | undefined = 5 * 1e9; // 5GB ceiling
@@ -122,6 +135,7 @@ class MockModelStore {
       contextId: computed,
       lastUsedModel: computed,
       activeModel: computed,
+      activeModelCaps: computed,
       displayModels: computed,
       availableModels: computed,
       isDownloading: computed,
@@ -240,6 +254,26 @@ class MockModelStore {
 
   get activeModel() {
     return this.models.find(model => model.id === this.activeModelId);
+  }
+
+  // Delegates to the real resolver over the live mock stores rather than
+  // returning a fixed answer, so consumer suites exercise the actual
+  // capability chain and stay reactive to store mutations.
+  private get capabilityEnv(): CapabilityEnv {
+    return {
+      remoteCaps: mockServerStore.remoteCaps,
+      binding: this.activeRemoteBinding,
+      isMultimodalActive: this.isMultimodalActive,
+      activeContextSettings: this.activeContextSettings,
+      activeModelId: this.activeModelId,
+    };
+  }
+
+  capsFor = (model: Model | undefined): ModelCapabilityView =>
+    resolveModelCaps(model, this.capabilityEnv);
+
+  get activeModelCaps(): ModelCapabilityView {
+    return this.capsFor(this.activeModel);
   }
 
   get displayModels(): Model[] {
--- a/__mocks__/stores/modelStore.ts
+++ b/__mocks__/stores/modelStore.ts
@@ -262,6 +262,7 @@ class MockModelStore {
   private get capabilityEnv(): CapabilityEnv {
     return {
       remoteCaps: mockServerStore.remoteCaps,
+      listCaps: mockServerStore.listCaps,
       binding: this.activeRemoteBinding,
       isMultimodalActive: this.isMultimodalActive,
       activeContextSettings: this.activeContextSettings,
--- a/__mocks__/stores/serverStore.ts
+++ b/__mocks__/stores/serverStore.ts
@@ -3,10 +3,20 @@ import {makeAutoObservable, observable} from 'mobx';
 import {RemoteModelCaps, ServerConfig} from '../../src/utils/types';
 import {ReasoningCapability} from '../../src/utils/reasoningCapability';
 import {RemoteModelInfo} from '../../src/api/openai';
+import {deriveListCapsMap} from '../../src/utils/listCaps';
 
 class MockServerStore {
   servers: ServerConfig[] = [];
   serverModels: Map<string, RemoteModelInfo[]> = observable.map();
+
+  // Derived from the live mock state, exactly as the real store derives it, so
+  // a suite that mutates `servers` or `serverModels` exercises the real
+  // derivation and stays reactive. A fixed answer here would let the card
+  // scenarios pass with the derivation broken.
+  get listCaps() {
+    return deriveListCapsMap(this.servers, this.serverModels);
+  }
+
   userSelectedModels: Array<{serverId: string; remoteModelId: string}> = [];
   remoteReasoning: Record<string, ReasoningCapability> = {};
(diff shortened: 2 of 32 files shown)


a-ghorbani:
## Visual evidence — per-model /props capabilities on a multi-model router







a-ghorbani:
## Visual evidence — model card capabilities













a-ghorbani:
## Visual evidence — capabilities read from the models list, before activation (iOS + Android)







































a-ghorbani:
## Visual evidence — add-sheet vision slot pinned to row edge (caption/icon no longer fuse)

Issue #625, opened 2026-03-09
Better Formatting and markdown support (Ui/Ux)


MohamedKhalilAssa:
It would be awesome and not that hard to add better markdown or formatting support in general. And it would make the app be easier on the eye in general.

Keep up the good work man!

a-ghorbani:
Any specific issues you're observing? We've made several formatting improvements recently, but it would be helpful if you could point to specific examples so we can look into them in a targeted way.

RedSystems586:
Yes, please add better formating, or at least an option to.

PR #834, opened 2026-07-24
fix: make markdown link color theme-aware (#753)

### LINKED ISSUE #753: [Bug]: URLs not distinguishable from regular text

sakebomb:
## Description

Links in model responses were indistinguishable from surrounding text — as reported in #753 ("rendered exactly the same as other text on the screen").

No user-agent tag style applies in this app (see the root cause below), so anchors get no link styling of their own and simply inherit the colour `MarkdownView` sets via `tagsStyles.body.color` (the theme's `text` token) — rendering as plain prose with no link affordance. Removing the `a` rule yields `#111111` in light and `#ffffff` in dark, the body text colour in both cases.

This binds the `<a>` tag style to the app's `secondary` theme token and adds an underline:

- Light: `#1E4DF6`
- Dark: `#95ABE6`

`marked` already autolinks bare URLs, `www.` URLs, and emails into `<a>` tags, so this covers every link form — not just explicit `[text](url)` markdown.

Fixes #753

### Correction to an earlier version of this description

This description has stated the mechanism wrongly **twice**, so both are recorded here rather than quietly edited out:

1. *"The renderer applies a hardcoded UA anchor colour of `#245dc1` and removing the `a` rule falls back to it."* Wrong — `UA_ANCHOR_COL` exists in `transient-render-engine` but never applies here, so it caused no part of the user-visible bug.
2. *"`body.color` takes precedence over the render engine's UA anchor styles."* Also wrong, and this is the wording still in commit `871dccb`. Measured, varying each axis independently (flattened style on the `<a>` node):

```
UA=off  body.color=on    {"color":"#111111"}                                    <- what users saw
UA=off  body.color=off   {}
UA=on   body.color=on    {"fontSize":14,"color":"#245dc1","textDecorationLine":"underline"}
UA=on   body.color=off   {"fontSize":14,"color":"#245dc1","textDecorationLine":"underline"}
```

With UA styles on, `#245dc1` wins *over* `body.color` — the reverse of the claim. There was never a precedence contest: no UA style was applying at all. Each of those two claims was written as a *correction* to the one before it, without being executed; correcting is not verifying.

Per @a-ghorbani's review, the underlying cause is broader than this PR: RNRH passes `enableUserAgentStyles: true` via `TRenderEngineProvider.defaultProps`, which React 19 ignores on function components, so **no** UA tag styles apply anywhere in the app — bold, italic, headings and lists all render as plain text too. This change is still wanted (the UA anchor colour isn't theme-aware), but it addresses links specifically rather than that root cause, which is being tracked separately.

## Platform Affected

- [x] iOS
- [x] Android

## Test plan

- [x] `Link Rendering` tests in `MarkdownView.test.tsx`: link text renders, links carry the theme secondary colour + underline, and a dark-mode case asserts the dark token.
- [x] Tests assert the theme token rather than a hex literal, so a palette retune doesn't fail on a bare colour.
- [x] **Mutation-checked** — removing the `a` rule turns both styling tests red (`#111111` light, `#ffffff` dark), confirming they guard the fix.
- [x] Verified `marked` autolinks all four forms (`https://…`, `www.…`, `a@b.com`, `[t](url)`) into `<a>` tags.
- [x] Full `MarkdownView` suite 21/21; full project suite 262 suites / 3952 tests. `tsc --noEmit` and eslint clean.
- [x] Rebased onto current `main`.
- [x] Dropped `textDecorationColor` (iOS-only, duplicated `color`) per review.
- [ ] Manual verification on a physical device — the reviewer captured before/after on an iOS simulator in both themes.

## Checklist

  - [ ] iOS Simulator/Device
  - [ ] Android Emulator/Device
- [x] Unit tests pass locally.

### DIFF
--- a/src/components/MarkdownView/__tests__/MarkdownView.test.tsx
+++ b/src/components/MarkdownView/__tests__/MarkdownView.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {ScrollView} from 'react-native';
+import {ScrollView, StyleSheet} from 'react-native';
 
 // Use the project's custom render which mounts MarkdownProvider — required
 // because MarkdownView now relies on the ambient TRenderEngineProvider
@@ -110,6 +110,33 @@ describe('MarkdownView Component', () => {
     );
 
     expect(getByText(/console\.log/)).toBeTruthy();
+  });
+
+  describe('Link Rendering', () => {
+    // Light theme secondary color (see src/theme/tokens/colors.ts).
+    const LINK_COLOR = '#1E4DF6';
+
+    it('renders link text from a markdown link', () => {
+      const markdownText = '[Example](https://example.com)';
+      const {getByText} = render(
+        <MarkdownView markdownText={markdownText} maxMessageWidth={300} />,
+      );
+
+      expect(getByText('Example')).toBeTruthy();
+    });
+
+    it('styles links with a distinct color and underline', () => {
+      const markdownText = '[Example](https://example.com)';
+      const {getByText} = render(
+        <MarkdownView markdownText={markdownText} maxMessageWidth={300} />,
+      );
+
+      const linkNode = getByText('Example');
+      const flattened = StyleSheet.flatten(linkNode.props.style) || {};
+
+      expect(flattened.color).toBe(LINK_COLOR);
+      expect(flattened.textDecorationLine).toBe('underline');
+    });
   });
 
   describe('Table Rendering', () => {
--- a/src/components/MarkdownView/__tests__/MarkdownView.test.tsx
+++ b/src/components/MarkdownView/__tests__/MarkdownView.test.tsx
@@ -5,6 +5,8 @@ import {ScrollView, StyleSheet} from 'react-native';
 // because MarkdownView now relies on the ambient TRenderEngineProvider
 // instead of building its own engine per instance.
 import {render, fireEvent} from '../../../../jest/test-utils';
+import {themeFixtures} from '../../../../jest/fixtures/theme';
+import {useTheme} from '../../../hooks';
 
 import {MarkdownView} from '../MarkdownView';
 
@@ -113,8 +115,8 @@ describe('MarkdownView Component', () => {
   });
 
   describe('Link Rendering', () => {
-    // Light theme secondary color (see src/theme/tokens/colors.ts).
-    const LINK_COLOR = '#1E4DF6';
+    const lightLink = themeFixtures.lightTheme.colors.secondary;
+    const darkLink = themeFixtures.darkTheme.colors.secondary;
 
     it('renders link text from a markdown link', () => {
       const markdownText = '[Example](https://example.com)';
@@ -134,8 +136,26 @@ describe('MarkdownView Component', () => {
       const linkNode = getByText('Example');
       const flattened = StyleSheet.flatten(linkNode.props.style) || {};
 
-      expect(flattened.color).toBe(LINK_COLOR);
+      expect(flattened.color).toBe(lightLink);
       expect(flattened.textDecorationLine).toBe('underline');
+    });
+
+    it('uses the dark theme link color in dark mode', () => {
+      (useTheme as jest.Mock).mockReturnValue(themeFixtures.darkTheme);
+      try {
+        const {getByText} = render(
+          <MarkdownView
+            markdownText="[Example](https://example.com)"
+            maxMessageWidth={300}
+          />,
+        );
+        const flattened =
+          StyleSheet.flatten(getByText('Example').props.style) || {};
+
+        expect(flattened.color).toBe(darkLink);
+      } finally {
+        (useTheme as jest.Mock).mockReturnValue(themeFixtures.lightTheme);
+      }
     });
   });
 
--- a/src/components/MarkdownView/styles.ts
+++ b/src/components/MarkdownView/styles.ts
@@ -21,14 +21,9 @@ export const createTagsStyles = (theme: Theme) => ({
     backgroundColor: 'transparent',
     // display: 'inline-block',
   },
-  // The `body` color above takes precedence over the render engine's UA anchor
-  // styles, so without an explicit rule links inherit the body text color and
-  // render indistinguishably from surrounding prose. Bind them to the theme's
-  // secondary color (plus an underline) so they read as links in both themes.
   a: {
     color: theme.colors.secondary,
     textDecorationLine: 'underline' as const,
-    textDecorationColor: theme.colors.secondary,
   },
   code: {
     fontFamily: 'Courier', // Change the font for code snippets
--- a/src/components/MarkdownView/styles.ts
+++ b/src/components/MarkdownView/styles.ts
@@ -21,6 +21,15 @@ export const createTagsStyles = (theme: Theme) => ({
     backgroundColor: 'transparent',
     // display: 'inline-block',
   },
+  // The `body` color above takes precedence over the render engine's UA anchor
+  // styles, so without an explicit rule links inherit the body text color and
+  // render indistinguishably from surrounding prose. Bind them to the theme's
+  // secondary color (plus an underline) so they read as links in both themes.
+  a: {
+    color: theme.colors.secondary,
+    textDecorationLine: 'underline' as const,
+    textDecorationColor: theme.colors.secondary,
+  },
   code: {
     fontFamily: 'Courier', // Change the font for code snippets
     backgroundColor: theme.colors.surface, // Custom background for code blocks


a-ghorbani:
## Visual evidence — markdown link styling, light + dark (before / after)

iPhone 17 Pro simulator (iOS 26.0), smollm2-135m. One chat turn containing all four link forms
`marked` produces — an explicit markdown link, a bare URL, a `www.` URL and an email — rendered in the
user bubble and echoed back in the assistant response. Same build, same prompt; only
`src/components/MarkdownView/styles.ts` differs between the two columns.

### Light

| before (`main`) | after (this PR) |
| --- | --- |
| light, before | light, after |

### Dark

| before (`main`) | after (this PR) |
| --- | --- |
| dark, before | dark, after |

Before the change every link form is indistinguishable from the surrounding prose — no colour, no
underline — in both themes. After it, all four render in `theme.colors.secondary` with an underline
(`#1E4DF6` light, `#95ABE6` dark): 6.1 : 1 contrast on the light chat background, 9.3 : 1 on the dark
one, both above WCAG AA.

src/components/MarkdownView/styles.ts
```diff
@@ -21,6 +21,15 @@ export const createTagsStyles = (theme: Theme) => ({
...
   },
+  // The `body` color above takes precedence over the render engine's UA anchor
+  // styles, so without an explicit rule links inherit the body text color and
+  // render indistinguishably from surrounding prose. Bind them to the theme's
+  // secondary color (plus an underline) so they read as links in both themes.
+  a: {
+    color: theme.colors.secondary,
+    textDecorationLine: 'underline' as const,
+    textDecorationColor: theme.colors.secondary,
+  },
```
a-ghorbani: Drop the comment. The code already says "link colour", and the stated reason isn't what happens: with this rule removed a link renders `#111111` (body colour, no underline), and removing `body.color` too still gives no colour and no underline — the UA anchor style never applies here. The real reason an explicit rule is needed is the React 19 / `defaultProps` issue in my review note, which is a property of the whole map rather than this entry.

`textDecorationColor` is iOS-only and repeats `color`, so it's dropped here too — say the word if you'd rather keep it.

```suggestion
  a: {
    color: theme.colors.secondary,
    textDecorationLine: 'underline' as const,
  },
```

src/components/MarkdownView/__tests__/MarkdownView.test.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
-import {ScrollView} from 'react-native';
+import {ScrollView, StyleSheet} from 'react-native';
```
a-ghorbani, line 2: Optional, pairs with the suggestion below.

```suggestion
import {ScrollView, StyleSheet} from 'react-native';

import {themeFixtures} from '../../../../jest/fixtures/theme';
import {useTheme} from '../../../hooks';
```

src/components/MarkdownView/__tests__/MarkdownView.test.tsx
```diff
@@ -112,6 +112,33 @@ describe('MarkdownView Component', () => {
...
+        <MarkdownView markdownText={markdownText} maxMessageWidth={300} />,
+      );
+
+      const linkNode = getByText('Example');
+      const flattened = StyleSheet.flatten(linkNode.props.style) || {};
+
+      expect(flattened.color).toBe(LINK_COLOR);
+      expect(flattened.textDecorationLine).toBe('underline');
+    });
+  });
```
a-ghorbani, line 160: Optional: assert the token rather than the literal, so a palette retune doesn't fail with a bare hex — the guard survives, since removing the `a` rule still yields `#111111`. And a dark case, which is the point of the PR: `useTheme` is mocked globally in `jest/setup.ts`, so pointing it at the dark fixture is enough.

```suggestion
  describe('Link Rendering', () => {
    const lightLink = themeFixtures.lightTheme.colors.secondary;
    const darkLink = themeFixtures.darkTheme.colors.secondary;

    it('renders link text from a markdown link', () => {
      const markdownText = '[Example](https://example.com)';
      const {getByText} = render(
        <MarkdownView markdownText={markdownText} maxMessageWidth={300} />,
      );

      expect(getByText('Example')).toBeTruthy();
    });

    it('styles links with a distinct color and underline', () => {
      const markdownText = '[Example](https://example.com)';
      const {getByText} = render(
        <MarkdownView markdownText={markdownText} maxMessageWidth={300} />,
      );

      const linkNode = getByText('Example');
      const flattened = StyleSheet.flatten(linkNode.props.style) || {};

      expect(flattened.color).toBe(lightLink);
      expect(flattened.textDecorationLine).toBe('underline');
    });

    it('uses the dark theme link color in dark mode', () => {
      (useTheme as jest.Mock).mockReturnValue(themeFixtures.darkTheme);
      try {
        const {getByText} = render(
          <MarkdownView
            markdownText="[Example](https://example.com)"
            maxMessageWidth={300}
          />,
        );
        const flattened =
          StyleSheet.flatten(getByText('Example').props.style) || {};

        expect(flattened.color).toBe(darkLink);
      } finally {
        (useTheme as jest.Mock).mockReturnValue(themeFixtures.lightTheme);
      }
    });
  });
```

a-ghorbani (changes requested):
The fix is right and works in both themes — before/after captures are in the comment above. lint, typecheck and the full jest suite pass; CI is green.

Two asks, both suggested inline (all three suggestions verified locally: lint, typecheck, 21/21 tests):

1. drop the comment above `a:` — see the inline note for why the stated reason doesn't hold.
2. the PR description says the renderer applies a hardcoded `#245dc1` and that removing the `a` style falls back to it. It doesn't — the mutation yields `#111111`. The commit message's wording ("links inherited the body text colour") is the accurate one.

The test suggestions (token instead of the literal, plus a dark-mode case) are optional.

**Separate bug, not yours** — I'll file it: no user-agent tag style applies anywhere in the app. RNRH passes `enableUserAgentStyles: true` through `TRenderEngineProvider.defaultProps`, and React 19 ignores `defaultProps` on function components, so the engine falls back to `false`. Only tags with an explicit `tagsStyles` entry are styled: `**bold**`, `*italic*`, headings and lists all render as plain text (visible in the captures, before and after). That's the root cause of #753, and probably #625. Your `a` rule is wanted either way, since the UA anchor colour isn't theme-aware.

sakebomb:
Thanks — both asks applied, and thank you for the device captures.

- **Comment dropped.** You're right that the reason didn't hold, and I should have caught it: I ran the mutation, saw `#111111`, and corrected the commit message on that basis — but left a comment asserting the `#245dc1` fallback anyway. The evidence was in front of me and the comment contradicted it.
- **PR description corrected**, with the wrong claim retracted explicitly rather than quietly edited out. Worth noting the first `gh pr edit` I ran appeared to fail with an unrelated Projects-classic GraphQL error and silently didn't apply — which is why the stale text was still there when you reviewed. Re-applied via the REST endpoint and verified.
- **`textDecorationColor` dropped** — no objection, iOS-only and duplicating `color`.
- **Both test suggestions taken**, not just as optional polish: the token assertion is strictly better, and the dark-mode case is the one that actually demonstrates the PR's premise. Removing the `a` rule now turns both red (`#111111` light, `#ffffff` dark).

On the `defaultProps` finding — that's a much better root cause than mine, and it explains something I saw but didn't follow up on: I'd noticed bold rendering flat in my own testing and assumed it was the simulator font. Thanks for filing it separately. Happy to pick it up if it's useful, or leave it with you.

sakebomb:
Following up on the `defaultProps` finding, since it hasn't been filed yet. I went looking to confirm it and ended up with a reproduction plus two corrections to the scope — recording it here so whoever writes the fix doesn't have to re-derive it.

**Confirmed, and it's reproducible in Jest — no device needed.** Rendering `plain **bolded**` under `MarkdownProvider` and flattening the style on the `bolded` node:

```
{"color":"#111111","fontSize":16,"fontFamily":"Helvetica Neue"}
```

No `fontWeight`. Adding `enableUserAgentStyles` to `TRenderEngineProvider` in `MarkdownProvider.tsx:58`:

```
{"fontSize":16,"color":"#111111","fontFamily":"Helvetica Neue","fontWeight":"bold"}
```

So the one-line fix does resolve it, and the symptom is unit-testable — a guard for this can live in the normal suite rather than resting on a screenshot.

**Correction 1 — the effective default is `false`, not "falls back to `false`".** I'd assumed the library default was `true` and React 19 was discarding it. Half right. `TRenderEngineProvider.defaultProps = defaultTRenderEngineProviderProps` at `TRenderEngineProvider.js:134` does set `enableUserAgentStyles: true` (`:85`), and React 19 does ignore it on a function component (`:119`). But the value doesn't become `undefined`-and-then-defaulted: `TRenderEngineProvider` spreads `...config` into `useTRenderEngine` → `buildTREFromConfig` → `new TRenderEngine({stylesConfig: {enableUserAgentStyles: undefined, ...}})`, and the engine's own `defaultStylesConfig` is `enableUserAgentStyles: **false**` (`@native-html/transient-render-engine/lib/commonjs/styles/defaults.js:10`). `createStylesConfig` only honours a passed value when `typeof === 'boolean'`, so `undefined` lands on the engine default of `false`. Net effect is identical, but it means this is the engine's documented default rather than a value being lost — which matters because it can't be fixed by a library patch that "restores defaultProps".

**Correction 2 — it's wider than `enableUserAgentStyles`.** Every default in that map is dropped, not just the one. Against what `MarkdownProvider.tsx:58-61` passes explicitly (`tagsStyles`, `customHTMLElementModels`, `systemFonts` — which is exactly why those three are the only things that work), the app is currently also losing:

| Dropped default | Effective value | Consequence |
| --- | --- | --- |
| `enableUserAgentStyles: true` | `false` (engine default) | bold/italic/headings/lists render flat — the #753 root cause |
| `emSize: 14` | `rootFontSize` falls back to `14` in `TRenderEngine` | no visible change today, but `em`/`rem` resolution now depends on a second fallback |
| `baseStyle: {fontSize: 14}` | engine's own `{fontSize: 14, whiteSpace: 'normal'}` | masked, but the app is no longer the one specifying it |
| `fallbackFonts` | `{}` in `buildTREFromConfig` | `isFontSupported` loses the ios/android sans-serif/monospace/serif map |
| `ignoredDomTags`, `ignoredStyles`, `classesStyles` | `undefined` / `[]` | benign here |

`enableCSSInlineProcessing` I initially listed as broken and it isn't worth flagging as a defect — the engine's own default is `true`, so it survives. `fallbackFonts` is the one besides UA styles with a real behavioural consequence.

**Suggested shape, if useful:** `enableUserAgentStyles` alone fixes #753, and is the minimal change. But since `defaultTRenderEngineProviderProps` is **not** exported from the package root (`index.d.ts:11` re-exports only `TRenderEngineProvider`, `defaultFallbackFonts` and `useAmbientTRenderEngine`), spreading the whole default map isn't available without a deep import. `defaultFallbackFonts` *is* exported, so `enableUserAgentStyles` + `fallbackFonts={defaultFallbackFonts}` covers both real gaps using only public API. Your call whether the font map is worth including or is better left to a separate change.

Two caveats on the above: this is all read from `node_modules` at `react-native-render-html@^6.3.4` / `react@19.1.1` as installed here, so it's the behaviour of the resolved tree rather than anything guaranteed by the library's docs. And the only thing I verified by execution is the bold `fontWeight` before/after — the rest of that table is source reading, not measurement, and I haven't checked any of it on a device.

Happy to take this as its own PR once there's an issue to attach it to, or to leave it entirely with you — you found it, and I don't want to step on it either way. Nothing here changes this PR: the `a` rule is still needed regardless, since the UA anchor colour isn't theme-aware.

sakebomb:
One more correction, and this one is on me twice over: the mechanism in this PR's commit message (`871dccb`) and description is also wrong.

Both say `tagsStyles.body.color` **took precedence over** the engine's UA anchor styles. It doesn't. I measured all four combinations — the `<a>` node's flattened style, varying UA styles and the `body.color` rule independently:

```
UA=off  body.color=on    {"color":"#111111"}                                    <- what users actually saw
UA=off  body.color=off   {}
UA=on   body.color=on    {"fontSize":14,"color":"#245dc1","textDecorationLine":"underline"}
UA=on   body.color=off   {"fontSize":14,"color":"#245dc1","textDecorationLine":"underline"}
```

With UA styles enabled, `#245dc1` wins *over* `body.color` — the opposite of what the commit message claims. The reason links rendered as prose was never precedence between the two; it's that no UA style was applying at all, which is your `defaultProps` finding. There was no competition to win.

Worth noting how this happened, since it's the more useful part: that wording was itself a *correction* to an earlier wrong claim (the `#245dc1` fallback). I replaced one unverified mechanism with another unverified mechanism and treated the act of correcting as if it were verification. A replacement inherits the original's unverified status — and by then this mechanism had been described three times without once being executed.

The whole thing was settleable in about twenty lines of throwaway Jest against the real component — no device, no simulator, available from the first round. I had the harness the entire time and reasoned instead.

**What doesn't change:** the fix. With UA styles on, an explicit `tagsStyles.a` still wins (`#1E4DF6` over `#245dc1`), so the rule is needed either way and is still the right change — which is what you said in your review. Only the stated reason was wrong, three times.

I'd rather not force-push a history rewrite over a message you've already reviewed. Happy to either reword `871dccb` and force-push, or leave the history and let the corrected mechanism live in the PR description and this thread — your call. I've corrected the description already.

sakebomb:
To close the loop on my own question above: I'll leave `871dccb`'s message as-is rather than force-push a reword. `f9191b2` sits on top of it, so rewording means rebasing two commits and force-pushing over a branch you've already reviewed — and the description plus this thread now record both wrong versions explicitly, which is better provenance than a rewritten message that hides them. Happy to reword if you'd rather the history stand alone; otherwise nothing further from me here.

Current state: description corrected, both review asks applied in `f9191b2`, CI green, no outstanding items on my side.

a-ghorbani (approved):
Both asks landed in `f9191b2` exactly as suggested, and the diff against `main` is now four lines of style plus the test block. Approving.

Re-verified locally on `f9191b2` (`origin/main` is an ancestor, so the rebase claim holds):

- `tsc --noEmit` clean, eslint clean on both changed files.
- `MarkdownView` 21/21; full suite 262 suites / 3952 passed, 310 snapshots — no snapshot drift from the `a` rule.
- **Mutation re-run**: deleting the `a` rule turns both styling tests red — `#111111` light, `#95ABE6` → `#ffffff` dark. The guard is real, and the dark case fails for a different reason than the light one, so it isn't decorative.

Two things I checked by execution rather than take on trust, both because they are the kind of claim this PR has already been bitten by:

**1. The affordance matches the behaviour.** Underlined blue text that doesn't open anything would be worse than unstyled text. The anchor node carries an `onPress`, and pressing it calls `Linking.openURL('https://example.com/')`. Worth recording *why* that survives when UA styles don't: the default `a.onPress` is merged in code — `mergeDeepRight(defaultRendererProps, props.renderersProps)` in `RenderersPropsProvider` — not delivered through `defaultProps`. Same library, two different default-delivery mechanisms; only the `defaultProps` one is broken under React 19.

**2. All four link forms actually carry the rule.** Not just that `marked` emits `<a>`, but that the style lands on each: explicit `[t](url)`, bare `https://`, `www.`, and a bare email all flatten to `{"color":"#1E4DF6", "textDecorationLine":"underline"}`.

**Contrast on the one background the captures don't cover.** My before/after captures measured the chat surface. A link pasted into the user's *own* message sits on `authorBubbleBackground` (`#f2f2f2` / `#212121`), not the surface — that's `Bubble/styles.ts`, which overrides the shell for the real chat. There it is 5.4 : 1 light and 7.1 : 1 dark, both above AA. And dropping `textDecorationColor` is visually inert — the underline falls back to the text colour, which was the same token — so the posted captures still stand for the final diff.

**On the commit-message question: nothing to do, and it isn't your call to make.** `main` squash-merges (#829 went in as 10 commits → one), so `871dccb`'s message never lands on `main` at all. The commit that does land takes its message from the PR title and body, which you've corrected. Leave the history alone — force-pushing would rewrite a message that was never going to be published anyway.

On the `defaultProps` follow-up: read, and useful — particularly the `undefined` → engine-default-`false` correction and `fallbackFonts`, which I hadn't traced. That work stays out of this PR; I'll come back to you separately on who picks it up rather than leave the offer hanging in a thread.

Issue #684, opened 2026-04-10
[bug]：The problem of model disappearance


TopFlowFox:
After I updated to version 1.13.3, the model disappeared. I upgraded from 1.13.1 to 1.13.3, and the memory was still occupied, so I had to delete the whole application and re-download it.

sakebomb:
A few questions that would help narrow this down before we keep going.

Was the model one you imported from your device files (using the "Add model" / file picker flow), or was it a preset model you downloaded from within the app, or a HuggingFace model you added by URL?

Also, what platform -- iOS or Android? And did you happen to do an iOS system update around the same time as the app upgrade?

The reason I ask: there are two different code paths here depending on the model type. For locally imported models on iOS, the app stores the absolute file path in its settings, and there are conditions where that path can become stale even though the file is still physically on disk. When the path check fails on startup, the app currently removes the model entry silently rather than flagging it as "file not found." That would explain the behavior you saw -- the file is still there taking up space, but the app has dropped its record of it.

If that is what happened, it is a bug worth fixing. The app should keep the entry and show a warning rather than deleting it without telling you.

sakebomb:
No reply on the questions above, so I went ahead with the fixes that hold regardless of the answers: #835.

To be clear about what that PR does and doesn't claim — I could not confirm your model type or platform, so I'm not asserting it's definitively the bug you hit. But your report (file still on disk, model gone from the list, right after an upgrade) is consistent with what I found:

For models imported from your device, the absolute file path is saved once at import time. On iOS the app container UUID can change during an OS upgrade or a backup/restore, which invalidates that saved path even though the file itself is untouched. The app then concluded the file was missing and removed the entry — and imported models are exactly the ones that can't be recovered, since there's no download URL to re-fetch from.

The PR re-anchors those stale paths onto the current container and repairs the saved entry, plus fixes two ordering bugs that made the check race its own results.

If you're still seeing this, the most useful thing would be whether the model was one you imported from your device files versus downloaded in-app — that distinguishes the two code paths.

PR #835, opened 2026-07-25
fix: stop local models disappearing when the app container path changes

sakebomb:
## Description

Imported local models are the only kind with no recovery path — there is no download URL to re-fetch from — yet they were the easiest to lose.

`fullPath` is persisted as an absolute path at import time. On iOS the app container UUID can change (OS upgrade, backup/restore, reinstall), which invalidates that path even though the file is still on disk under the new container. The stale path then failed its existence check, the model was marked not-downloaded, and `removeInvalidLocalModels` deleted the entry outright.

Three separate defects contributed:

1. **`refreshDownloadStatuses` never awaited its per-model checks.** It used `forEach` over an async function, so it resolved immediately with every check still in flight. Callers that depend on `isDownloaded` being settled — notably `initializeStore`, which calls `removeInvalidLocalModels` on the very next line — raced them. Each check is now awaited *and* isolated, so one failure neither aborts the others nor records an indeterminate result as "file is gone".
2. **`mergeModelLists` dropped the promise** from `initializeDownloadStatus`, so even the `await`ed migration path did not actually wait for it. It now returns that promise (kept sync-callable, since `resetModels` invokes it inside `runInAction`).
3. **Nothing attempted to recover a stale path.** Imported models are copied into `Documents/models/local` (see `handleAddLocalModel`), so that suffix is stable and can be re-anchored onto the *current* documents path. When the file is found there, the entry is repaired and the corrected path persisted so later launches skip the recovery.

Removing a local model is now logged rather than silent.

The iOS Shortcuts resolver (`PalDataProvider.swift:parseModelPath`) duplicates this logic — as the note on `getModelFullPath` requires — and had the identical staleness bug, so it gets the matching recovery.

Refs #684

### Scope note

I could not get the reporter on #684 to confirm model type or platform, so I am **not** claiming this is definitively their bug — the report is consistent with it but unconfirmed. Defects 1 and 2 are order-of-operations bugs that stand on their own regardless. Behavior when the file is genuinely gone is unchanged: the model is still removed.

### Blast radius (recovery)

- **iOS-only in effect.** Android's `DocumentDirectoryPath` is `filesDir`, which is stable across upgrades/reinstalls, so no local path goes stale there. Defects 1–2 still affect both platforms; the re-anchoring only ever fires on iOS.
- **Pre-`8244a688` (Nov 2024) imports** predate the `models/local` destination, carry no marker, and are intentionally left unrecovered (`markerIndex === -1` returns them unchanged) — they're still pruned if their file is gone.
- The `/models/local/` marker is now a behavioural invariant shared with the Swift resolver; both sides carry a sync warning.

### Review follow-up (commit `4b913ea`)

Addresses maintainer review:

- **`resolveLocalModelPath` no longer lets an FS rejection escape.** Both `exists()` checks are wrapped and degrade to "not present", keeping `getModelFullPath` total for `deleteModel` (which awaits it outside its `try`) and its other call sites.
- **The migration branch now prunes too.** `removeInvalidLocalModels()` runs after the awaited `mergeModelLists`, making the "settle then prune" comment honest; safe because the status checks are settled first.
- **Recovered paths are batched.** One `runInAction` flushed after the refresh pass replaces one full store serialization per recovered model.
- **Guard tests strengthened.** The failing-check test now seeds `isDownloaded: true` and asserts survival, plus a pruning-survival case (12 tests total). Mutation-verified — injecting the forbidden `isDownloaded = false` into the refresh `catch` turns both red.

## Platform Affected

- [x] iOS
- [x] Android

(Defects 1–2 affect both; the container-path staleness and the Swift change are iOS-specific.)

## Test plan

New suite `src/store/__tests__/ModelStore.localModelPath.test.ts` (10 tests) covering path recovery, the await contract, and pruning.

- [x] **Mutation-checked every fix** — each reverts to a red test:
  - revert `resolveLocalModelPath` call site → "keeps a local model whose stale path was recovered" fails
  - restore the `forEach` → unhandled rejection crashes the run (exit 1); with the fix, exit 0
  - drop the `return` in `mergeModelLists` → "returns a promise that settles the status checks" fails
- [x] **Suite fails on base, passes with the fix** — on `4f1ba9b` the new file is the only failing suite (262/263 pass); with the change 263/263.
- [x] Full suite: 263 suites / 3959 tests pass. `tsc --noEmit`, eslint, prettier clean.
- [x] Verified path re-anchoring against nested subdirectories and non-`models/local` paths (left untouched).
- [x] Confirmed the pre-existing "worker process failed to exit gracefully" warning also occurs on base — not introduced here.
- [x] **Swift change compiles** — confirmed by the maintainer in review (`pod install` + iOS/Android builds pass, `PalDataProvider.o` in the main target). A Shortcuts smoke test on a real recovered path is still outstanding. I avoided `appendingPathComponent` on a leading-`/` component.
- [ ] Manual device verification of the container-change scenario (needs a real backup/restore or reinstall).

## Checklist

  - [ ] iOS Simulator/Device
  - [ ] Android Emulator/Device

### DIFF
--- a/ios/PocketPal/AppIntents/PalDataProvider.swift
+++ b/ios/PocketPal/AppIntents/PalDataProvider.swift
@@ -197,6 +197,27 @@ class PalDataProvider {
                 print("[PalDataProvider] Error: Full path is undefined for local model")
                 return nil
             }
+            if fileManager.fileExists(atPath: fullPath) {
+                return fullPath
+            }
+            // The stored path is absolute and captured at import time; the app
+            // container UUID can change (OS upgrade, backup/restore), leaving it
+            // stale while the file still exists under the current container.
+            // Imported models are copied into Documents/models/local, so
+            // re-anchor that stable suffix onto the current documents path.
+            // Mirrors resolveLocalModelPath() in src/store/ModelStore.ts.
+            // Use upperBound and re-add the known prefix so the appended
+            // component never begins with "/".
+            if let markerRange = fullPath.range(of: "/models/local/") {
+                let relativePath = "models/local/" + String(fullPath[markerRange.upperBound...])
+                let recoveredPath = documentsPath
+                    .appendingPathComponent(relativePath)
+                    .path
+                if recoveredPath != fullPath, fileManager.fileExists(atPath: recoveredPath) {
+                    print("[PalDataProvider] Recovered local model path: \(recoveredPath)")
+                    return recoveredPath
+                }
+            }
             return fullPath
         }
 
--- a/src/store/ModelStore.ts
+++ b/src/store/ModelStore.ts
@@ -545,7 +545,7 @@ class ModelStore {
     const presets = await this.resolvePresets();
 
     if (storedVersion < MODEL_LIST_VERSION) {
-      this.mergeModelLists(presets);
+      await this.mergeModelLists(presets);
       // Only finalize the one-time migration once presets actually resolved.
       // An empty result signals a transient resolve failure (e.g. the RAM read
       // rejected); leave the version unbumped so the migration retries next
@@ -818,7 +818,9 @@ class ModelStore {
 
     this.reconcilePresets(presets);
 
-    this.initializeDownloadStatus();
+    // Returned (not awaited) so callers inside runInAction can stay sync while
+    // initializeStore can still await settlement before pruning local models.
+    return this.initializeDownloadStatus();
   };
 
   setupAppStateListener = () => {
@@ -970,7 +972,7 @@ class ModelStore {
       if (!model.fullPath) {
         throw new Error('Full path is undefined for local model');
       }
-      return model.fullPath;
+      return this.resolveLocalModelPath(model);
     }
 
     if (!model.filename) {
@@ -1048,6 +1050,51 @@ class ModelStore {
     return `${RNFS.DocumentDirectoryPath}/${model.filename}`;
   };
 
+  /**
+   * Re-anchors a LOCAL model's stored absolute path onto the current app
+   * container.
+   *
+   * `fullPath` is persisted verbatim when the file is imported. On iOS the app
+   * container UUID can change (OS upgrade, backup/restore, reinstall), which
+   * invalidates every stored absolute path even though the file is still on
+   * disk under the new container. Imported models are copied into
+   * `<DocumentDirectoryPath>/models/local`, so the segment from `models/local`
+   * onward is stable and can be re-joined to the *current* documents path.
+   *
+   * Returns the recovered path (and persists it) when the original is missing
+   * but the re-anchored one exists; otherwise returns the stored path unchanged
+   * so existing not-found handling still applies.
+   */
+  resolveLocalModelPath = async (model: Model): Promise<string> => {
+    const storedPath = model.fullPath!;
+
+    if (await RNFS.exists(storedPath)) {
+      return storedPath;
+    }
+
+    const marker = '/models/local/';
+    const markerIndex = storedPath.indexOf(marker);
+    if (markerIndex === -1) {
+      return storedPath;
+    }
+
+    const recoveredPath = `${RNFS.DocumentDirectoryPath}${storedPath.slice(
+      markerIndex,
+    )}`;
+    if (recoveredPath === storedPath || !(await RNFS.exists(recoveredPath))) {
+      return storedPath;
+    }
+
+    console.log(
+      '[ModelStore] Recovered local model path after container change:',
+      {modelId: model.id, from: storedPath, to: recoveredPath},
+    );
+    runInAction(() => {
+      model.fullPath = recoveredPath;
+    });
+    return recoveredPath;
+  };
+
   async checkFileExists(model: Model) {
     const filePath = await this.getModelFullPath(model);
     const exists = await RNFS.exists(filePath);
@@ -1071,9 +1118,25 @@ class ModelStore {
   }
 
   refreshDownloadStatuses = async () => {
-    this.models.forEach(model => {
-      this.checkFileExists(model);
-    });
+    // Await every check: callers (notably initializeStore, which then runs
+    // removeInvalidLocalModels) depend on isDownloaded being settled. A bare
+    // forEach resolved immediately and left the checks in flight.
+    //
+    // Each check is isolated: a single failure (e.g. an RNFS error, or a local
+    // model with no fullPath) must not abort the others, and must leave
+    // isDownloaded untouched rather than defaulting it to false — an
+    // indeterminate result is not evidence the file is gone.
+    await Promise.all(
+      this.models.map(model =>
+        this.checkFileExists(model).catch(error => {
+          console.warn(
+            '[ModelStore] Could not determine file status for model:',
+            model.id,
+            error,
+          );
+        }),
+      ),
+    );
   };
 
   initializeDownloadStatus = async () => {
@@ -1082,14 +1145,26 @@ class ModelStore {
 
   removeInvalidLocalModels = () => {
     runInAction(() => {
-      this.models = this.models.filter(
-        model =>
-          // Keep all non-local models (preset and HF)
-          !(model.isLocal || model.origin === ModelOrigin.LOCAL) ||
-          // This condition ensures that we keep models that are downloaded.
-          // For local models, isDownloaded==true means the file exists, otherwise it's invalid.
-          model.isDownloaded,
-      );
+      this.models = this.models.filter(model => {
+        // Keep all non-local models (preset and HF)
+        if (!(model.isLocal || model.origin === ModelOrigin.LOCAL)) {
+          return true;
+        }
+        // For local models, isDownloaded==true means the file exists, otherwise
+        // it's invalid. Callers must let refreshDownloadStatuses settle first —
+        // it re-anchors stale container paths (resolveLocalModelPath) so a
+        // recoverable model is not treated as missing here.
+        if (model.isDownloaded) {
+          return true;
+        }
+        // Dropping a local model is unrecoverable (there is no download URL to
+        // re-fetch from), so make it visible rather than silent.
+        console.warn(
+          '[ModelStore] Removing local model whose file is no longer present:',
+          {modelId: model.id, name: model.name, fullPath: model.fullPath},
+        );
+        return false;
+      });
     });
   };
 
--- a/src/store/ModelStore.ts
+++ b/src/store/ModelStore.ts
@@ -1079,7 +1079,7 @@ class ModelStore {
    * existing not-found handling still applies.
    *
    * IMPORTANT: this re-anchoring is duplicated in native Swift for iOS
-   * Shortcuts — see the LOCAL branch of resolveModelPath() in
+   * Shortcuts — see the LOCAL branch of parseModelPath() in
    * ios/PocketPal/AppIntents/PalDataProvider.swift. Keep the two in sync; the
    * shared '/models/local/' marker must match on both sides.
    */
@@ -1108,10 +1108,6 @@ class ModelStore {
       return storedPath;
     }
 
-    // Imported models are copied under <Documents>/models/local, so the segment
-    // from that marker onward is stable across container-UUID churn (iOS OS
-    // upgrade, backup/restore, reinstall) and can be re-joined to the current
-    // documents path.
     const marker = '/models/local/';
     const markerIndex = storedPath.indexOf(marker);
     if (markerIndex === -1) {
--- a/src/store/ModelStore.ts
+++ b/src/store/ModelStore.ts
@@ -197,6 +197,13 @@ class ModelStore {
   // Last requested model ID - enables "last one wins" during rapid switching
   private pendingModelId: string | null = null;
 
+  // Recovered local-model paths awaiting persistence. resolveLocalModelPath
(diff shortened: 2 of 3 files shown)


src/store/__tests__/ModelStore.localModelPath.test.ts
```diff
@@ -0,0 +1,240 @@
...
+          return Promise.reject(new Error('RNFS blew up'));
+        }
+        return Promise.resolve(path === healthyPath);
+      });
+
+      await expect(
+        modelStore.refreshDownloadStatuses(),
+      ).resolves.toBeUndefined();
+
+      expect(modelStore.models[1].isDownloaded).toBe(true);
```
a-ghorbani, line 187: **This test can't fail.**

The guarantee at `ModelStore.ts:1125-1128` — an indeterminate result isn't evidence the file is gone — is the load-bearing claim of the whole fix, and it's asserted in prose only.

The broken model is seeded `isDownloaded: false` on line 162 and then never asserted on. The only assertions are `resolves.toBeUndefined()` and `models[1]`. Put `model.isDownloaded = false` inside the `.catch` in `refreshDownloadStatuses` — the exact regression the comment forbids — and this still passes.

Seed it `true`, assert it stays `true`, and add a pruning case proving it survives `removeInvalidLocalModels`.

src/store/ModelStore.ts
```diff
@@ -1048,6 +1050,51 @@ class ModelStore {
...
+   * onward is stable and can be re-joined to the *current* documents path.
+   *
+   * Returns the recovered path (and persists it) when the original is missing
+   * but the re-anchored one exists; otherwise returns the stored path unchanged
+   * so existing not-found handling still applies.
+   */
+  resolveLocalModelPath = async (model: Model): Promise<string> => {
+    const storedPath = model.fullPath!;
+
+    if (await RNFS.exists(storedPath)) {
```
a-ghorbani: **An FS rejection can now escape `getModelFullPath`.**

Every other `exists` in this function is wrapped and falls through to a canonical path (`:997`, `:1006`, `:1035`). The LOCAL branch used to be total — `return model.fullPath` can't throw. Now it can, and this one and the one at `:1084` are both unguarded.

The consumer that matters is `deleteModel:1378`, which does `await this.getModelFullPath(_model)` **outside** its `try` — a rejection aborts the delete before the splice and before the unlink. `updateModelHash:3489` is invoked as a floating promise, so there it's an unhandled rejection.

Not theoretical — `@dr.pogodin/react-native-fs`'s Android `exists` has an explicit reject path:

```kotlin
override fun exists(filepath: String, promise: Promise) {
    try { promise.resolve(File(filepath).exists()) }
    catch (ex: Exception) { ex.printStackTrace(); reject(promise, filepath, ex) }
}
```

Wrap both `exists` calls and degrade to `false`. Keeps the "indeterminate isn't absence" intent without turning a path lookup into a throwing operation for eleven call sites.

src/store/ModelStore.ts
```diff
@@ -818,7 +818,9 @@ class ModelStore {
 
     this.reconcilePresets(presets);
 
-    this.initializeDownloadStatus();
+    // Returned (not awaited) so callers inside runInAction can stay sync while
+    // initializeStore can still await settlement before pruning local models.
```
a-ghorbani, line 834: This comment describes a call path that doesn't exist — `mergeModelLists` is only reached from the migration branch, and that branch never prunes. Details in the review summary.

src/store/ModelStore.ts
```diff
@@ -1048,6 +1050,51 @@ class ModelStore {
...
+    )}`;
+    if (recoveredPath === storedPath || !(await RNFS.exists(recoveredPath))) {
+      return storedPath;
+    }
+
+    console.log(
+      '[ModelStore] Recovered local model path after container change:',
+      {modelId: model.id, from: storedPath, to: recoveredPath},
+    );
+    runInAction(() => {
```
a-ghorbani, line 1147: **Each recovery triggers a full store serialization.**

Every recovered path gets its own `runInAction`, and `mobx-persist-store` is watching `models` with no debounce (`configurePersistable` is never called). So each one is a full `toJS` deep-clone + `JSON.stringify` + `AsyncStorage.setItem` of the entire models array, on the JS thread, during startup.

A container change invalidates every local path at once — so this is N full serialize+write cycles where one would do.

Return the recovered path without mutating, collect the pairs, and apply them in a single `runInAction` after `Promise.all` settles.

src/store/ModelStore.ts
```diff
@@ -1048,6 +1050,51 @@ class ModelStore {
...
+   * so existing not-found handling still applies.
+   */
+  resolveLocalModelPath = async (model: Model): Promise<string> => {
+    const storedPath = model.fullPath!;
+
+    if (await RNFS.exists(storedPath)) {
+      return storedPath;
+    }
+
+    const marker = '/models/local/';
```
a-ghorbani, line 1111: Models imported before `8244a688` (Nov 2024) predate the `models/local` destination, so they carry no marker and get no recovery — `markerIndex === -1` returns them unchanged and they're still pruned.

Fine as a limit, but worth a line in the PR about the intended blast radius.

Separately: this marker is now a behavioural invariant living as four literals across three files and two languages (here, `ModelsScreen.tsx:200`, and twice in `PalDataProvider.swift:211-212`), spelled inconsistently — the writer has no trailing slash, this reader requires one. A missed site fails silently.

src/store/ModelStore.ts
```diff
@@ -1071,9 +1118,25 @@ class ModelStore {
   }
 
   refreshDownloadStatuses = async () => {
-    this.models.forEach(model => {
-      this.checkFileExists(model);
-    });
+    // Await every check: callers (notably initializeStore, which then runs
+    // removeInvalidLocalModels) depend on isDownloaded being settled. A bare
```
a-ghorbani: **Comment load** — ~40 comment lines against ~67 production lines in this diff.

This one narrates the bug you just fixed ("a bare `forEach` resolved immediately and left the checks in flight"). That's task history; it stops being meaningful the moment the old code is gone. `:1160-1161` restates the `console.warn` directly below it. `:1125-1128` exists because the isolation logic is inline — extract a `checkFileExistsIsolated` and the prose goes away with it.

The two genuine "why"s here — container UUID churn, and the leading-`/` trap in `appendingPathComponent` — are worth about four lines. Those are the ones to keep.

ios/PocketPal/AppIntents/PalDataProvider.swift
```diff
@@ -197,6 +197,27 @@ class PalDataProvider {
...
             }
+            if fileManager.fileExists(atPath: fullPath) {
+                return fullPath
+            }
+            // The stored path is absolute and captured at import time; the app
+            // container UUID can change (OS upgrade, backup/restore), leaving it
+            // stale while the file still exists under the current container.
+            // Imported models are copied into Documents/models/local, so
+            // re-anchor that stable suffix onto the current documents path.
+            // Mirrors resolveLocalModelPath() in src/store/ModelStore.ts.
```
a-ghorbani, line 208: The `IMPORTANT: duplicated in native Swift` warning lives on `getModelFullPath` (`ModelStore.ts:961-963`) and is scoped to "this function". The code you actually mirrored is `resolveLocalModelPath`, which carries no such note — so the new duplication escapes the one mechanism keeping these two in sync. Worth putting the warning on the function that's now mirrored, in both directions.

Two pre-existing nits on lines right above: `:169` cites `getModelFullPath() (lines ~618-659)`, ~350 lines off (it's at `:969`), and that whole `///` block is syntactically attached to `inferRepoFromModelId`, not to `parseModelPath`, which has no doc comment at all.

One asymmetry worth knowing: this resolver reads `local_pals.default_model`, a frozen snapshot that's exported and imported between devices (`exportUtils.ts:214` / `importUtils.ts:435`). So an imported Pal can carry a foreign device's absolute path. Before this change that path was dead and the Shortcut failed cleanly; now it can re-anchor onto a same-named local file the Pal was never authored against. The TS side is immune because it resolves by `id`. Gating the re-anchor on the stored prefix ending in `/Documents` would narrow it.

a-ghorbani (commented):
Reviewed this. Diagnosis is right and both ordering defects are real — nice find on the `forEach`.

First, the thing you flagged as unrun: **the Swift compiles.** I ran it here — `pod install`, iOS build, Android build all pass, and `PalDataProvider.o` lands in the main app target for both arm64 and x86_64. Also `tsc`/eslint/prettier clean, 263 suites / 3959 tests pass, `ModelStore.ts` at 79.7% stmts / 68.7% branch. That checkbox is covered; the Shortcuts smoke test on a real recovered path still isn't.

Specifics are inline. Three things that don't sit on changed lines:

**The migration branch still never prunes.** The comment at `:821-823` says the promise is returned so `initializeStore` can await settlement "before pruning local models", but `mergeModelLists` is only reached from the migration branch, which never calls `removeInvalidLocalModels`. The branch that does prune (`:559-561`) skips `mergeModelLists` entirely and already awaited on base. `resetModels:2600` now drops the promise on the floor.

The missing prune is pre-existing, and your `Promise.all` is what makes fixing it safe — on base, pruning there would have run against last session's persisted flags. So either add the prune (which makes the `await` load-bearing) or drop the `await` and the comment.

**`:559-561` is uncovered.** That's the only production site of "settle, then prune" — the thing the whole fix depends on. The tests hand-roll the sequence in the test body instead of driving `initializeStore`. Reorder or drop the prune in production and nothing goes red.

**On the #684 link — worth narrowing.** `backup_rules_12_plus.xml` excludes `models/` from both `<cloud-backup>` and `<device-transfer>` while the model list itself is backed up. An Android restore therefore gives "list restored, files actually gone", which this correctly still prunes. If the reporter is on Android, this isn't their fix. Android's `DocumentDirectoryPath` is `filesDir` and stable anyway, so the recovery is iOS-only in effect.

---

One direction question for the maintainer, not you: `model.fullPath` is the only persisted absolute path in the codebase — `imageUtils.ts` and the PRESET/HF branches all recompute from `DocumentDirectoryPath`, which is exactly why only LOCAL has this bug. Persisting `models/local/<file>` and joining at read time would make this class of bug impossible and delete the Swift marker search entirely. Bigger than a bug fix, but worth deciding whether this is the interim or the end state.

Security pass came back clean, for the record: the re-anchor can only move a path into the sandbox, `..` was already in the persisted string, and the rewrite is gated on `exists`. `indexOf`/`range(of:)` first-occurrence is the correct choice here — last-occurrence would've been the buggy one. TS/Swift parity checked with a compiled harness over 9 inputs, matches byte-for-byte including the doubled-marker case.

sakebomb:
Thanks — the FS-guard and the migration-prune catches were both real. Addressed in `4b913ea`.

**Fixed**
- **FS rejection escaping `getModelFullPath`.** Both `exists()` calls in `resolveLocalModelPath` are now wrapped and degrade an indeterminate result to "not present", so the function stays total for its eleven call sites and `deleteModel`'s pre-splice `await` can't abort on an RNFS throw.
- **Migration branch never pruned.** Added `removeInvalidLocalModels()` after the awaited `mergeModelLists` in the `storedVersion < MODEL_LIST_VERSION` branch — this is the option that makes the `await` load-bearing rather than dropping it, and it's safe precisely because the status checks are now settled first (as you noted). The comment now matches the code.
- **N serializations per container change.** `resolveLocalModelPath` collects `(model → recovered path)` and a single `flushPendingLocalPathRecoveries()` after `refreshDownloadStatuses` settles applies them in one `runInAction`. Callers use the returned path immediately, so nothing observes the deferral.
- **The test that couldn't fail.** Seeded the broken model `isDownloaded: true`, assert it survives, and added a pruning-survival case. Mutation-verified: injecting `model.isDownloaded = false` into the refresh `catch` turns both red.
- **Swift sync warning.** Put an `IMPORTANT` note on `resolveLocalModelPath` pointing at the mirrored `resolveModelPath()` LOCAL branch, so the duplication is flagged in both directions (the Swift side already pointed back here).

Verification: `tsc --noEmit` clean, `eslint` clean, `ModelStore` suites 212/212 green. No native changes in this commit, so the iOS/Android builds you already ran still stand.

**Deferred, for your call**
- **Android blast radius** — agreed the recovery is iOS-only in effect (`filesDir` is stable), and pre-`8244a688` imports carry no marker and are still pruned. I'll fold both into the PR description.
- **Imported-Pal re-anchor asymmetry** (foreign absolute path re-anchoring onto a same-named local file via the Shortcut) — this is Swift-only and I can't compile/verify it here the way you can. If you'd like the `…/Documents` prefix guard, I'll add it in a follow-up; flagging rather than blind-editing native code.
- **Relative-path persistence** — that's the bigger direction question you raised for the maintainer; happy to prototype it separately if it's the intended end state.

a-ghorbani (approved):
Approving. The diagnosis was right, both ordering defects were real, and the follow-up commit addressed every blocking point from the review.

Verified locally on 8b146fd7: tsc/eslint/prettier clean, full jest 263 suites / 3961 tests, pod install, iOS build, Android build, and iOS e2e quick-smoke. The error-isolation guarantee is mutation-checked — injecting the forbidden regression turns the intended tests red.

One caveat recorded honestly: Android e2e quick-smoke fails, but it fails identically on origin/main (5e0f72b5) against the same emulator and spec, so it is pre-existing and not from this change. Tracking separately.

Thanks for the thorough test plan and the accurate self-assessment on what you couldn't verify.

PR #836, opened 2026-07-25
feat: export a chat session as Markdown

sakebomb:
## Description

Implements #845.

The existing JSON export is built for re-importing — it isn't something you can read, paste into a doc, or send to someone who doesn't use the app. This adds a second export format alongside it, in the same Export/Import menu: **Export current session as Markdown**.

Output is a `.md` file with the session title, an export timestamp, and `### User` / `### Assistant` headings per message.

Two details worth calling out:

- **Message content goes through the existing `toExportedMessage` projection**, not `message.text` directly. `assistant_turn` rows have an empty `text` column by design, so reading it directly would export blank assistant messages — the projection routes through `derivedText()` for the visible joined step content, matching what the JSON export already does.
- **Author attribution keys off the author id**, not message position, so it stays correct when a Pal greeting means the assistant speaks first.

`shareJsonData` gains an optional `mimeType` defaulting to `application/json`, so every existing caller is unchanged.

Only the English string is added; other locales fall back to English until Weblate picks it up.

No new dependencies and no native code.

## Platform Affected

- [x] iOS
- [x] Android

## Test plan

Six new tests in `src/utils/__tests__/exportUtils.test.ts`.

- [x] **Mutation-checked all three behaviors** — each reverts to a red test:
  - read `msg.text` instead of the projection → the `assistant_turn` test fails (this is the bug the projection avoids)
  - un-thread `mimeType` → the markdown mime-type test fails
  - attribute role by position instead of author id → the rendering test fails
- [x] Existing export tests unchanged and passing (35/35 in that suite), including the JSON export still sharing as `application/json`.
- [x] Full suite: 262 suites / 3955 tests pass. `tsc --noEmit` clean; `npm run lint` reports 0 errors (7 pre-existing warnings).
- [x] `HeaderRight` suite passes (8/8) with the new menu item.
- [x] Kept the `en.json` diff to a single line — note that file already fails `prettier --check` on `main`, so I did not run `--write` on it (it would have reformatted an unrelated `byteSizes` block).
- [ ] Manual device verification of the share sheet on iOS/Android.

## Checklist

  - [ ] iOS Simulator/Device
  - [ ] Android Emulator/Device

### DIFF
--- a/src/components/HeaderRight/HeaderRight.tsx
+++ b/src/components/HeaderRight/HeaderRight.tsx
@@ -27,6 +27,7 @@ import {t} from '../../locales';
 import {importChatSessions} from '../../utils/importUtils';
 import {
   exportChatSession,
+  exportChatSessionAsMarkdown,
   exportAllChatSessions,
 } from '../../utils/exportUtils';
 
@@ -108,6 +109,18 @@ export const HeaderRight: React.FC = observer(() => {
         await exportChatSession(session.id);
       } catch (error) {
         console.error('Error exporting current session:', error);
+        Alert.alert('Export Error', 'Failed to export the current session.');
+      }
+    }
+    closeMenu();
+  };
+
+  const onPressExportCurrentSessionAsMarkdown = async () => {
+    if (session?.id) {
+      try {
+        await exportChatSessionAsMarkdown(session.id);
+      } catch (error) {
+        console.error('Error exporting current session as markdown:', error);
         Alert.alert('Export Error', 'Failed to export the current session.');
       }
     }
@@ -237,6 +250,12 @@ export const HeaderRight: React.FC = observer(() => {
               key="export-current"
               onPress={onPressExportCurrentSession}
               label={l10n.components.headerRight.exportCurrentSession}
+            />,
+            <Menu.Item
+              disabled={!session?.id}
+              key="export-current-markdown"
+              onPress={onPressExportCurrentSessionAsMarkdown}
+              label={l10n.components.headerRight.exportCurrentSessionMarkdown}
             />,
             <Menu.Item
               key="export-all"
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -749,7 +749,7 @@
       "makeChatTemporary": "Make chat temporary",
       "export": "Export/Import",
       "exportCurrentSession": "Export current session",
-      "exportCurrentSessionMarkdown": "Export current session as Markdown",
+      "exportCurrentSessionMarkdown": "Export as Markdown",
       "exportAllSessions": "Export all sessions",
       "exportChatSession": "Export chat session",
       "importSessions": "Import sessions"
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -749,6 +749,7 @@
       "makeChatTemporary": "Make chat temporary",
       "export": "Export/Import",
       "exportCurrentSession": "Export current session",
+      "exportCurrentSessionMarkdown": "Export current session as Markdown",
       "exportAllSessions": "Export all sessions",
       "exportChatSession": "Export chat session",
       "importSessions": "Import sessions"
--- a/src/utils/__tests__/exportUtils.test.ts
+++ b/src/utils/__tests__/exportUtils.test.ts
@@ -20,6 +20,7 @@ import {
   exportChatSessionAsMarkdown,
 } from '../exportUtils';
 import {userId} from '../chat';
+import {format} from 'date-fns';
 import {ensureLegacyStoragePermission} from '../androidPermission';
 
 // Mock dependencies
@@ -48,6 +49,9 @@ jest.mock('@dr.pogodin/react-native-fs', () => ({
 }));
 
 jest.mock('date-fns', () => ({
+  // Keep the fixed return value every existing test relies on for filenames,
+  // but stay input-aware so a test can assert WHICH date was formatted. A
+  // return-only mock makes "export time vs session date" unobservable.
   format: jest.fn().mockReturnValue('2024-01-01_12-00-00'),
 }));
 
@@ -627,23 +631,27 @@ describe('exportUtils', () => {
   });
 
   describe('exportChatSessionAsMarkdown', () => {
+    // `createdAt` ascends with the argument order so fixtures can be written
+    // newest-first (see the ordering note on the repository mock below) while
+    // still carrying truthful timestamps.
     const makeTextMessage = (
       id: string,
       author: string,
       text: string,
+      createdAt = 1704067200000,
     ): any => ({
       id,
       author,
       text,
       type: 'text',
       metadata: null,
-      createdAt: 1704067200000,
+      createdAt,
       toMessageObject: () => ({
         id,
         type: 'text',
         text,
         author: {id: author},
-        createdAt: 1704067200000,
+        createdAt,
         metadata: {},
       }),
     });
@@ -654,6 +662,11 @@ describe('exportUtils', () => {
     };
 
     beforeEach(() => {
+      // IMPORTANT: `getSessionById` queries `Q.sortBy('position', Q.desc)` and
+      // `addMessageToSession` assigns `highestPosition + 1`, so the real array
+      // is NEWEST FIRST — the same convention ChatSessionStore documents
+      // ("messages are in reverse order, ie 0 is the latest"). Fixtures must
+      // match that order or an ordering bug stays invisible.
       chatSessionRepository.getSessionById = jest.fn().mockResolvedValue({
         session: {
           id: 'session-1',
@@ -662,8 +675,8 @@ describe('exportUtils', () => {
           activePalId: null,
         },
         messages: [
-          makeTextMessage('m1', userId, 'What is 2+2?'),
-          makeTextMessage('m2', 'assistant-1', 'It is 4.'),
+          makeTextMessage('m2', 'assistant-1', 'It is 4.', 1704067260000),
+          makeTextMessage('m1', userId, 'What is 2+2?', 1704067200000),
         ],
         completionSettings: {settings: '{}'},
       } as any);
@@ -690,6 +703,51 @@ describe('exportUtils', () => {
       expect(content).toContain('It is 4.');
     });
 
+    it('writes the transcript in chronological order, oldest first', async () => {
+      await exportChatSessionAsMarkdown('session-1');
+
+      const {content} = writtenMarkdown();
+      // Positional, not membership: `toContain` alone passes in any order.
+      expect(content.indexOf('What is 2+2?')).toBeLessThan(
+        content.indexOf('It is 4.'),
+      );
+      expect(content.match(/^### (User|Assistant)$/gm)).toEqual([
+        '### User',
+        '### Assistant',
+      ]);
+    });
+
+    it('stamps the export time, not the session creation date', async () => {
+      // `session.date` is only ever set at creation and never refreshed, so a
+      // chat started years ago must not claim to have been exported then.
+      const sessionCreated = '2020-06-15T08:30:00Z';
+      (chatSessionRepository.getSessionById as jest.Mock).mockResolvedValueOnce(
+        {
+          session: {
+            id: 'session-old',
+            title: 'Old Chat',
+            date: sessionCreated,
+            activePalId: null,
+          },
+          messages: [makeTextMessage('m1', userId, 'Hi')],
+          completionSettings: {settings: '{}'},
+        } as any,
+      );
+
+      await exportChatSessionAsMarkdown('session-old');
+
+      // `format` is mocked to a fixed string, so the rendered output cannot
+      // reveal which date was used — assert on the argument instead.
+      const formattedDates = (format as jest.Mock).mock.calls
+        .filter(([, pattern]) => pattern === 'yyyy-MM-dd HH:mm')
+        .map(([date]) => (date as Date).toISOString());
+
+      expect(formattedDates).toHaveLength(1);
+      expect(formattedDates[0]).not.toBe(
+        new Date(sessionCreated).toISOString(),
+      );
+    });
+
     it('attributes authorship by user id rather than message order', async () => {
       // Assistant speaks first here; role must follow the author id.
       (chatSessionRepository.getSessionById as jest.Mock).mockResolvedValueOnce(
@@ -700,9 +758,11 @@ describe('exportUtils', () => {
             date: '2024-01-01T00:00:00Z',
             activePalId: null,
           },
+          // Newest first: the Pal's greeting is the OLDEST message here, so it
+          // sits last in the repository's array and first in the transcript.
           messages: [
-            makeTextMessage('m1', 'assistant-1', 'Hi there!'),
-            makeTextMessage('m2', userId, 'Hello.'),
+            makeTextMessage('m2', userId, 'Hello.', 1704067260000),
+            makeTextMessage('m1', 'assistant-1', 'Hi there!', 1704067200000),
           ],
           completionSettings: {settings: '{}'},
         } as any,
@@ -711,9 +771,15 @@ describe('exportUtils', () => {
       await exportChatSessionAsMarkdown('session-2');
 
       const {content} = writtenMarkdown();
-      expect(content.indexOf('### Assistant')).toBeLessThan(
-        content.indexOf('### User'),
+      // A Pal greeting means the assistant legitimately speaks first; role must
+      // follow the author id, and the greeting must lead the transcript.
+      expect(content.indexOf('Hi there!')).toBeLessThan(
+        content.indexOf('Hello.'),
       );
+      expect(content.match(/^### (User|Assistant)$/gm)).toEqual([
+        '### Assistant',
+        '### User',
+      ]);
     });
 
     it('exports assistant_turn content, whose text column is empty by design', async () => {
--- a/src/utils/__tests__/exportUtils.test.ts
+++ b/src/utils/__tests__/exportUtils.test.ts
@@ -49,9 +49,10 @@ jest.mock('@dr.pogodin/react-native-fs', () => ({
 }));
 
 jest.mock('date-fns', () => ({
-  // Keep the fixed return value every existing test relies on for filenames,
-  // but stay input-aware so a test can assert WHICH date was formatted. A
-  // return-only mock makes "export time vs session date" unobservable.
+  // Return-only mock with the fixed value the filename tests rely on. Tests
+  // that need to assert WHICH date was formatted (export time vs session date)
+  // read `format.mock.calls` — `jest.fn()` records arguments regardless of
+  // return value.
   format: jest.fn().mockReturnValue('2024-01-01_12-00-00'),
 }));
 
@@ -866,12 +867,16 @@ describe('exportUtils', () => {
       await exportChatSessionAsMarkdown('session-5');
 
       const {content} = writtenMarkdown();
-      expect(content).toContain('_[image: file:///a.png]_');
-      expect(content).toContain('_[image: file:///b.png]_');
+      // One note per attachment, none dropped silently.
+      const noteCount = content.split('_[image]_').length - 1;
+      expect(noteCount).toBe(2);
       // The note follows the text it belongs to, inside the same section.
       expect(content.indexOf('Look at this')).toBeLessThan(
-        content.indexOf('_[image: file:///a.png]_'),
+        content.indexOf('_[image]_'),
       );
+      // The on-device temp URI must never leak into a shared transcript.
+      expect(content).not.toContain('file:///a.png');
+      expect(content).not.toContain('file:///b.png');
     });
 
     it('throws when the session does not exist', async () => {
--- a/src/utils/__tests__/exportUtils.test.ts
+++ b/src/utils/__tests__/exportUtils.test.ts
@@ -820,6 +820,60 @@ describe('exportUtils', () => {
       expect(content).toContain('Second part.');
     });
 
+    it('collapses a multi-line title so it stays a single heading', async () => {
+      // Titles are derived from the user's first message, so they can contain
+      // newlines — and a `#` starting the second line shifts the structure.
+      (chatSessionRepository.getSessionById as jest.Mock).mockResolvedValueOnce(
+        {
+          session: {
+            id: 'session-4',
+            title: 'Multi\nline # title',
+            date: '2024-01-01T00:00:00Z',
+            activePalId: null,
+          },
+          messages: [makeTextMessage('m1', userId, 'Hi')],
+          completionSettings: {settings: '{}'},
+        } as any,
+      );
+
+      await exportChatSessionAsMarkdown('session-4');
+
+      const {content} = writtenMarkdown();
+      expect(content).toContain('# Multi line # title');
+      expect(content.match(/^#[^#]/gm)).toHaveLength(1);
+    });
+
+    it('notes attached images instead of dropping them silently', async () => {
+      const withImages = makeTextMessage('m1', userId, 'Look at this');
+      // `toExportedMessage` reads the raw `metadata` column, not the in-memory
+      // object, so the URIs have to be on the JSON string.
+      withImages.metadata = JSON.stringify({
+        imageUris: ['file:///a.png', 'file:///b.png'],
+      });
+      (chatSessionRepository.getSessionById as jest.Mock).mockResolvedValueOnce(
+        {
+          session: {
+            id: 'session-5',
+            title: 'With Images',
+            date: '2024-01-01T00:00:00Z',
+            activePalId: null,
+          },
+          messages: [withImages],
+          completionSettings: {settings: '{}'},
+        } as any,
+      );
+
+      await exportChatSessionAsMarkdown('session-5');
+
+      const {content} = writtenMarkdown();
+      expect(content).toContain('_[image: file:///a.png]_');
+      expect(content).toContain('_[image: file:///b.png]_');
+      // The note follows the text it belongs to, inside the same section.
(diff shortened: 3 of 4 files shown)


src/utils/exportUtils.ts
```diff
@@ -271,11 +271,60 @@ export const exportLegacyChatSessions = async (): Promise<void> => {
...
+    const lines: string[] = [
+      `# ${session.title}`,
+      '',
+      `*Exported: ${exportedAt}*`,
+      '',
+      '---',
+      '',
+    ];
+
+    for (const msg of messages) {
```
a-ghorbani: **Blocking — the exported transcript comes out in reverse chronological order.**

This iterates `getSessionById().messages` in the order the repository returns it, and that order is newest-first, not chronological:

- `ChatSessionRepository.getSessionById` sorts `Q.sortBy('position', Q.desc)`
- `addMessageToSession` assigns `record.position = highestPosition + 1`, so the newest message always has the highest `position`
- therefore `position DESC` == newest first

`ChatSessionStore` states the same convention explicitly — it `unshift`s new messages, with the comment "messages are in reverse order, ie 0 is the latest" — and the chat screen renders through an inverted `FlatList`, which is why it looks right on screen.

So every exported file puts each answer *above* the question that prompted it, oldest message last:

```
### Assistant
ANSWER 2 (newest)
### User
QUESTION 2
### Assistant
ANSWER 1
### User
QUESTION 1 (oldest)
```

Confirmed two ways rather than inferred:

1. A real export written by the existing JSON path from a genuine chat session has the assistant reply at index 0 and the user prompt at index 1, `createdAt` strictly **descending**. That is the array this loop walks.
2. Feeding a fixture in that same real order produces the heading sequence `["Assistant","User","Assistant","User"]` with the oldest message at the end of the file.

The JSON export is unaffected because it is a machine format and `createSession` re-inverts on import (`position = initialMessages.length - i`) — newest-first is an internal convention that round-trips consistently. Markdown is the first consumer for which the order is human-facing.

```suggestion
    for (const msg of [...messages].reverse()) {
```

(Copy rather than `messages.reverse()` — the array is the repository's result and shouldn't be mutated in place.)

src/utils/exportUtils.ts
```diff
@@ -271,11 +271,60 @@ export const exportLegacyChatSessions = async (): Promise<void> => {
...
+  sessionId: string,
+): Promise<void> => {
+  try {
+    const sessionData = await chatSessionRepository.getSessionById(sessionId);
+    if (!sessionData) {
+      throw new Error('Session not found');
+    }
+
+    const {session, messages} = sessionData;
+    const exportedAt = format(new Date(session.date), 'yyyy-MM-dd HH:mm');
```
a-ghorbani: **The line rendered as `*Exported: …*` is actually the session's creation date.**

`session.date` is written in exactly two places, neither of which is an update:

- `createSession` — `record.date = new Date().toISOString()`
- the legacy JSON→WatermelonDB migration — copies the old value verbatim

Nothing refreshes it afterwards, so it isn't even "last activity" — it is creation, permanently. A chat started in January and exported in July renders `*Exported: <January>*`.

Reproduced on an Android device: a session dated `2026-01-15`, exported on `2026-07-26`, wrote `*Exported: 2026-01-15 10:30*` into a file named `chat_..._2026-07-26_10-01-40.md`.

Related, same line: `date-fns` v4 `format` throws `RangeError: Invalid time value` on an unparseable date, which would reject the whole Markdown export while the JSON export of the same session still succeeds. Low likelihood, and the fix below removes it for free.

If the intent is the export time (which is what the label and the PR description both say):

```suggestion
    const exportedAt = format(new Date(), 'yyyy-MM-dd HH:mm');
```

If the session's own date is genuinely the more useful thing for a reader of a shared transcript — a defensible product call — then relabel the line to `*Session date: …*` instead and add a parse guard, since that path keeps calling `format` on stored data of unvalidated shape. Either way one of the two has to change.

src/locales/en.json
```diff
@@ -749,6 +749,7 @@
       "makeChatTemporary": "Make chat temporary",
       "export": "Export/Import",
       "exportCurrentSession": "Export current session",
+      "exportCurrentSessionMarkdown": "Export current session as Markdown",
```
a-ghorbani: **This label truncates at accessibility font sizes, and the truncation lands on the word "Markdown".**

Measured on an Android 16 device. At default font scale the string renders at 732 px against 452 / 373 / 317 px for its three siblings — roughly twice the longest existing item in the menu. The submenu auto-widens to fit, so it never wraps; it ellipsizes once the menu can grow no further, which happens between `font_scale` 1.3 and 1.5:

| `font_scale` | rendered |
| --- | --- |
| 1.0 | `Export current session as Markdown` |
| 1.3 | `Export current session as Markdown` |
| 1.5 | `Export current session as Markdo…` |
| 2.0 | `Export current session as Markdo…` |

truncated label

Full screen at `font_scale` 1.5:

truncated menu

That cut removes the only thing distinguishing this item from `Export current session` directly above it, so at large text sizes two adjacent items read almost identically. Every sibling stays fully readable at the same scale, so this is specific to the new string. `font_scale` 1.5–2.0 is reachable through Android's Accessibility → Display size and text.

It also compounds for translations: since the English source already truncates at 1.5, a longer translation truncates sooner — and every non-`en` locale currently falls back to this English string.

Suggested shortening — the item already sits under an `Export/Import` parent with `Export current session` directly above, so the scope reads from context, and this touches only the new key (no existing source string changes, so no existing translation is marked stale):

```suggestion
      "exportCurrentSessionMarkdown": "Export as Markdown",
```

Considered and set aside: a symmetric `(JSON)` / `(MD)` pair reads well but abbreviates to jargon and lengthens the currently-safe sibling from 452 px to ~609 px, regressing *its* truncation threshold; dropping the redundant `Export` prefix from all three items is the nicest information design but re-translates three existing keys.

Separately, and not something to fix here: the menu never tells the user the existing export is JSON. Worth addressing on its own where the translation churn can be paid deliberately.

src/components/HeaderRight/HeaderRight.tsx
```diff
@@ -238,6 +251,12 @@ export const HeaderRight: React.FC = observer(() => {
               onPress={onPressExportCurrentSession}
               label={l10n.components.headerRight.exportCurrentSession}
             />,
+            <Menu.Item
+              disabled={!session?.id}
+              key="export-current-markdown"
```
a-ghorbani, line 256: **Visual evidence for the new menu item** (captured on Android 16, API 36, with this branch's JS).

Active session — full submenu, all items enabled:

export submenu, session active

No active session — both export-current items correctly greyed out, which confirms the `disabled={!session?.id}` guard. Note the existing `Export current session` item gains the same guard here, so this is also a small fix to a pre-existing gap:

export submenu, no session

The item renders on a single line at default font scale with about 6% width headroom. See the note on the label string for what happens at larger font scales.

src/utils/exportUtils.ts
```diff
@@ -271,11 +271,60 @@ export const exportLegacyChatSessions = async (): Promise<void> => {
...
+      lines.push(`### ${role}`, '', exported.text ?? '', '', '---', '');
+    }
+
+    const timestamp = format(new Date(), 'yyyy-MM-dd_HH-mm-ss');
+    const sanitizedTitle = session.title
+      .replace(/[^a-z0-9]/gi, '_')
+      .toLowerCase();
+    const filename = `chat_${sanitizedTitle}_${timestamp}.md`;
+
+    await shareJsonData(lines.join('\n'), filename, 'text/markdown');
```
a-ghorbani, line 333: **Checked this on a device — no issue, recording the result so it doesn't get re-litigated.**

Threading a new MIME type into `Share.open` looked like it might shrink the share sheet, since Android matches `ACTION_SEND` targets by MIME. Measured on an Android 16 image with `cmd package query-activities`:

| MIME | resolvers |
| --- | --- |
| `text/plain` | 8 |
| `text/markdown` | 3 |
| `application/json` | 3 |

`text/markdown` and `application/json` resolve to the **identical** package set (Drive, Gmail, GMS/Quick Share), and the live sheet matched exactly:

share sheet for the exported .md

Since `application/json` is what every existing export already sends, this introduces no regression in share reach and no dead-end. The default-parameter approach also leaves all existing callers provably unchanged.

Worth knowing for later: `text/plain` would reach 8 targets on the same image, adding Messages, Chrome and Bluetooth, and the `.md` extension would still carry the format for anything that opens the file. That's an improvement to the whole export path rather than anything this PR broke.

End-to-end on device the export works: file written to Downloads, well-formed Markdown, correct role attribution, inline formatting preserved. The only defect in the output is the message ordering flagged above.

src/utils/__tests__/exportUtils.test.ts
```diff
@@ -623,4 +625,152 @@ describe('exportUtils', () => {
...
+      }),
+    });
+
+    const writtenMarkdown = () => {
+      const call = (RNFS.writeFile as jest.Mock).mock.calls.at(-1);
+      return {path: call?.[0] as string, content: call?.[1] as string};
+    };
+
+    beforeEach(() => {
+      chatSessionRepository.getSessionById = jest.fn().mockResolvedValue({
```
a-ghorbani, line 671: **This fixture is ordered the opposite way from the repository it stands in for**, which is why a fully reversed transcript passes all six new tests.

`getSessionById` returns `Q.sortBy('position', Q.desc)` — newest first. This fixture is `[user 'What is 2+2?', assistant 'It is 4.']`, i.e. oldest first, so the real repository would hand these back in the opposite order.

Worth ordering the fixture newest-first to match the real contract, and asserting the *rendered* sequence is chronological (oldest heading first) rather than just that both headings are present. A small shared helper that encodes the repository's ordering once would stop the next consumer of `getSessionById` having to rediscover it.

For context, the pre-existing JSON-export tests in this file are not affected — `exportChatSession`'s fixture holds a single message and `exportAllChatSessions`' holds `messages: []`, so neither exercised ordering in either direction. The gap is new because Markdown is the first consumer for which order is user-visible.

src/utils/__tests__/exportUtils.test.ts
```diff
@@ -623,4 +625,152 @@ describe('exportUtils', () => {
...
+
+      const {content} = writtenMarkdown();
+      expect(content).toContain('# My Chat');
+      expect(content).toContain('### User');
+      expect(content).toContain('What is 2+2?');
+      expect(content).toContain('### Assistant');
+      expect(content).toContain('It is 4.');
+    });
+
+    it('attributes authorship by user id rather than message order', async () => {
```
a-ghorbani, line 752: **This test currently locks in the reversed output.**

The fixture is `[assistant 'Hi there!', user 'Hello.']` and the assertion is that `### Assistant` precedes `### User`. But under the repository's real newest-first ordering, `[assistant, user]` is exactly what an ordinary user-asks / assistant-answers exchange looks like coming *out* of `getSessionById` — and the correct chronological rendering of that is User first, then Assistant. So the assertion passes on reversed output.

It also ends up contradicting its own comment: a genuine Pal greeting, where the assistant really does speak first in time, would arrive from the repository as `[user 'Hello.', assistant 'Hi there!']`.

The intent behind the test is good and worth keeping — author attribution should follow the author id, not position. Writing the fixture the way the repository would actually return it, and asserting the rendered order is chronological, would keep that intent while letting the ordering bug fail loudly.

src/utils/__tests__/exportUtils.test.ts
```diff
@@ -623,4 +625,152 @@ describe('exportUtils', () => {
...
+      await exportChatSessionAsMarkdown('session-1');
+
+      const {path} = writtenMarkdown();
+      expect(path).toMatch(/\.md$/);
+      expect(Share.open).toHaveBeenCalledWith(
+        expect.objectContaining({type: 'text/markdown'}),
+      );
+    });
+
+    it('renders the title, headings and message text as markdown', async () => {
```
a-ghorbani, line 696: Small coverage gap: this asserts the title, both headings and both message bodies, but nothing asserts the `*Exported: …*` line — which is the one line with a defect in it.

The fixture already supplies a fixed `date: '2024-01-01T00:00:00Z'`, so once the timestamp fix lands, faking the clock and asserting that line would be cheap. Keeping a fixture whose `session.date` differs from "now" is what keeps the two apart permanently.

src/utils/exportUtils.ts
```diff
@@ -271,11 +271,60 @@ export const exportLegacyChatSessions = async (): Promise<void> => {
 };
 
 /**
- * Helper function to share JSON data as a file
+ * Export a single chat session as a human-readable Markdown file.
```
a-ghorbani, line 274: Minor: the second paragraph here restates what `toExportedMessage`'s own docstring already says, about thirty lines up, at the code that implements it. Two copies of one rationale tend to drift, and the copy at the call site is the one that rots because it describes someone else's invariant.

The first sentence is worth keeping and isn't duplicated anywhere — "the JSON export is for re-importing; this one is for reading and sharing outside the app" is a genuine reason for having two export formats that a reader can't derive from either function body.

src/utils/exportUtils.ts
```diff
@@ -271,11 +271,60 @@ export const exportLegacyChatSessions = async (): Promise<void> => {
...
+    const sessionData = await chatSessionRepository.getSessionById(sessionId);
+    if (!sessionData) {
+      throw new Error('Session not found');
+    }
+
+    const {session, messages} = sessionData;
+    const exportedAt = format(new Date(session.date), 'yyyy-MM-dd HH:mm');
+
+    const lines: string[] = [
+      `# ${session.title}`,
```
a-ghorbani: Minor: `session.title` is interpolated raw, and titles are generated from the user's own message text (`ChatSessionStore` takes `derivedText(message)` and truncates it), so a multi-line first message yields a multi-line title.

Verified: a title of `Multi\nline # title` renders as an H1 `# Multi` followed by `line # title` as a body line, and a `#` or `-` starting that second line shifts the structure further. Cosmetic — it's the user's own content in their own file — but `session.title.replace(/\s+/g, ' ').trim()` would settle it.

src/utils/exportUtils.ts
```diff
@@ -271,11 +271,60 @@ export const exportLegacyChatSessions = async (): Promise<void> => {
...
+      `*Exported: ${exportedAt}*`,
+      '',
+      '---',
+      '',
+    ];
+
+    for (const msg of messages) {
+      const exported = toExportedMessage(msg);
+      const role = exported.author === userId ? 'User' : 'Assistant';
+      lines.push(`### ${role}`, '', exported.text ?? '', '', '---', '');
```
a-ghorbani: Minor, worth a thought rather than a change request: attached images are dropped with no trace.

Multimodal messages carry their images in `metadata.imageUris`, and `msg.toMessageObject()` already lifts them onto the in-memory message, so they're available right here. A message that had images exports as bare text with nothing indicating an attachment existed — and the JSON export at least round-trips the URIs.

Sending an image requires non-empty text, so this never produces an empty section; only the images go missing. A text-only first version is a perfectly reasonable scope call, it just shouldn't be silent — one `image` or an `_[image]_` placeholder per entry would do it.

src/components/HeaderRight/HeaderRight.tsx
```diff
@@ -114,6 +115,18 @@ export const HeaderRight: React.FC = observer(() => {
     closeMenu();
   };
 
+  const onPressExportCurrentSessionAsMarkdown = async () => {
+    if (session?.id) {
+      try {
+        await exportChatSessionAsMarkdown(session.id);
+      } catch (error) {
+        console.error('Error exporting current session as markdown:', error);
```
a-ghorbani, line 123: Minor, and pre-existing rather than introduced here: this produces two stacked alerts on a share failure.

`shareJsonData` already shows a localized `exportError` alert and *then* rethrows, so by the time this `catch` runs the user has seen one alert, and this adds a second — hardcoded English, against the convention that new UI strings go through l10n.

`onPressExportCurrentSession` just above has exactly the same two lines, so the JSON export already behaves this way and this mirrors it faithfully. Flagging it so the duplication is a deliberate choice rather than inherited by copy-paste — worth fixing in both places or neither. Dropping the handler-level alert (the util already reports to the user) or reusing `l10n.components.exportUtils.exportError` would both work; a handler-level alert does still have a job for the paths that never reach `shareJsonData`, like `Session not found`.

a-ghorbani (changes requested):
Reviewed on an Android 16 device with this branch's JS running live, plus the full JS suite, typecheck and lint. Nice PR — the subtle part, routing content through the existing `toExportedMessage` projection rather than `message.text`, is exactly the right call.

**Blocking:** the exported transcript comes out in reverse chronological order for any real session. Detail and a one-line fix are on `exportUtils.ts`; the test fixtures need reordering with it, or the suite keeps passing over the bug.

Everything else is inline: the `*Exported:*` timestamp, the menu label truncating at large font scales, three test notes and four minors. One thing I checked and cleared rather than raised — `text/markdown` resolves to the same share targets as the `application/json` already in use — is recorded on the line so it doesn't get re-litigated.

**Verification:** typecheck clean; lint 0 errors (7 pre-existing warnings, none in changed files); 262 suites / 3955 tests green; changed-surface coverage 80% statements, above the 60% floor. `NATIVE_CHANGES=NO`, so no pod install or native builds. Visual evidence embedded inline. iOS not captured. Your `en.json` prettier note is accurate — it fails identically on `main`, so leaving it alone was right.

---

sakebomb:
Good catch on the ordering — that was a real bug and it's fixed in 749e3d4.

Applied the `[...messages].reverse()` suggestion verbatim, including the copy rather than in-place reverse.

**Why my tests didn't catch it**, since that's the more useful part: my fixtures were hand-written oldest-first, because that's how a conversation reads to a human. I never checked what `getSessionById` actually returns. Compounding it, every assertion was `toContain`, which is order-blind — so the suite would have stayed green under any permutation. I'd mutation-tested all three behaviours and all three went red, which gave me false confidence: mutation testing validates the code under test, and the defect was in my assumption about the data. Those are orthogonal, and I treated passing one as evidence for the other.

Fixtures now match the real `position DESC` order with a comment explaining why, and two tests assert relative position. Reverting the fix turns both red.

Also applied:

- **`*Exported:*` now stamps the export time.** You're right that `session.date` is creation-only. Verifying this needed the `format` mock's *arguments* — the mock returns a fixed string, so the rendered output can't reveal which date was passed. Worth flagging: the existing mock is return-only, so any future test asserting *which* date got formatted will silently pass. This also removes the `RangeError` path you mentioned.
- **Label shortened to "Export as Markdown."** Your reasoning on the alternatives is convincing — particularly that the `(JSON)`/`(MD)` pair would regress the currently-safe sibling. Agreed the menu not disclosing that the existing export is JSON is a separate concern with its own translation cost.

Thanks for recording the `text/markdown` share-target check rather than raising it — that would otherwise have been re-litigated, and I hadn't verified it myself.

One thing I should flag on process: I opened this PR without filing an issue first, which CONTRIBUTING asks for so the feature can be discussed in advance. Happy to file one retroactively, or to close this if a human-readable export isn't wanted — the review effort here shouldn't be taken as commitment to merge.

sakebomb:
Three of the four minors are fixed in 961b5df. The double-alert one I've left alone deliberately — reasoning below.

**Images noted rather than dropped** (`exportUtils.ts:308`). You were right that the silence was the problem, not the scope: the JSON export round-trips the URIs, so a Markdown transcript that loses them without a word is worse than one that admits to it. Each attachment gets an `_[image: <uri>]_` note inside its own section. I went with the note over `image` on purpose — the URIs point at on-device app storage, so a real embed renders broken everywhere the transcript actually gets read, which is a worse outcome than a visible placeholder. Happy to switch if you'd rather have the embed.

**Multi-line title collapsed** (`exportUtils.ts:294`). Applied `replace(/\s+/g, ' ').trim()`. Reproduced your case first — `Multi\nline # title` did render as an H1 `# Multi` plus a body line.

**Docstring trimmed** (`exportUtils.ts:274`). Dropped the duplicated paragraph and kept the first sentence, for exactly the reason you gave: the call-site copy is the one that rots.

**Not fixed: the double alert** (`HeaderRight.tsx:123`). Your framing — "worth fixing in both places or neither" — is what decided it. Fixing it properly means touching `onPressExportCurrentSession`, which is pre-existing code this PR otherwise doesn't go near, and it'd widen the diff on a PR whose merge isn't settled yet. It's also a genuinely separate defect: `shareJsonData` both alerting *and* rethrowing is the actual design problem, and it affects every export path, not just the new one. I'd rather do it as its own change where the two handlers can be fixed together and the hardcoded English replaced with `l10n.components.exportUtils.exportError` in both. Say the word if you'd prefer it here and I'll fold it in.

**Verification:** typecheck clean; eslint and prettier clean on both changed files; full suite 262/262 suites, 3959 passed / 2 skipped. Both new behaviours mutation-guarded — dropping the title collapse or the image loop turns the new tests red. One caveat worth stating plainly: I hit a single `BenchmarkScreen` failure on my first full run and couldn't reproduce it in three subsequent full runs or in isolation (45/45). Its stack trace lands in `ChatView.test.tsx` via `react-native-image-viewing`'s deprecated `SafeAreaView`, and nothing under `BenchmarkScreen` or `ChatView` imports `exportUtils`, so I'm calling it an unrelated flake rather than something I introduced — but I didn't chase it to a root cause. No device verification on this round; the image-note path in particular has only been exercised in tests, not against a real multimodal message.

src/utils/exportUtils.ts
```diff
@@ -271,11 +271,79 @@ export const exportLegacyChatSessions = async (): Promise<void> => {
...
+      lines.push(`### ${role}`, '', exported.text ?? '');
+
+      // Multimodal messages keep their attachments in `metadata.imageUris`.
+      // Note the URI rather than embedding it: these point at on-device app
+      // storage, so an `![...]()` embed would render broken everywhere the
+      // transcript is actually read.
+      const imageUris: unknown = exported.metadata?.imageUris;
+      if (Array.isArray(imageUris)) {
+        for (const uri of imageUris) {
+          lines.push('', `_[image: ${uri}]_`);
```
a-ghorbani: Minor: noting the attachment was the right response — silence was the problem. But the URI itself is dead on arrival.

These come straight from `react-native-image-picker` and are never copied anywhere persistent:

- Android — `Utils.java`: `getCacheDir()`, prefix `rn_image_picker_lib_temp_`, with the library's own comment "getCacheDir will auto-clean according to android docs"
- iOS — `ImagePickerManager.mm`: `NSTemporaryDirectory()`

So the line points at an OS-evictable temp file inside app-private storage. It resolves for no reader, often won't resolve on the originating device either, and on iOS the path carries the app-install container UUID into a document whose whole purpose is being shared.

You asked whether I'd rather have `image` — neither. You're right that an embed renders broken; the fix is to drop the URI, not to link it. `_[image]_`, or just the basename, carries the same "an attachment was here" signal without the noise or the UUID.

(Checked the underscores in `rn_image_picker_lib_temp_` don't break the italics — CommonMark ignores intraword `_`. Not an issue.)

src/utils/__tests__/exportUtils.test.ts
```diff
@@ -46,6 +49,9 @@ jest.mock('@dr.pogodin/react-native-fs', () => ({
 }));
 
 jest.mock('date-fns', () => ({
+  // Keep the fixed return value every existing test relies on for filenames,
```
a-ghorbani: Trivial: this comment describes something the code doesn't do.

It says the mock stays "input-aware" and that "a return-only mock makes this unobservable" — but the mock is still `jest.fn().mockReturnValue(...)`, i.e. return-only. The observability your new test relies on comes from `jest.fn()` recording calls, which was already true before this diff. Nothing here changed.

The underlying point is worth keeping, just as the reason the test asserts on `mock.calls` instead of the output.

a-ghorbani (commented):
All eleven items from the last round are closed. I mutation-tested the four code fixes rather than reading them — reverting each one turns the suite red, so the new tests are real guards:

| reverted | result |
| --- | --- |
| `[...messages].reverse()` → `messages` | 2 fail |
| `new Date()` → `new Date(session.date)` | 1 fail |
| title collapse | 1 fail |
| image-note loop | 1 fail |

A rendered sample export (fixture in the repository's real `position DESC` order) now comes out chronological with the export date stamped. Asserting the `format` **arguments** for the timestamp was the right call — the return-only mock can't reveal it any other way.

Leaving the double alert alone is fine by me. My framing was "both places or neither", you chose neither consistently, and the real defect is `shareJsonData` alerting *and* rethrowing — which belongs in its own change.

Two new minors inline, one worth acting on. Neither blocks.

**Verification:** typecheck clean; lint 0 errors (7 pre-existing warnings, none in changed files); 262 suites / 3959 passed / 2 skipped. Your `BenchmarkScreen` flake didn't reproduce here. `NATIVE_CHANGES=NO`. No device this round, unlike last time — so nothing here is device-measured.

---

src/utils/exportUtils.ts
sakebomb: Agreed — dropped the URI in cfeb63c, the note now reads `_[image]_`.

Went with `_[image]_` over the basename because on Android the basename *is* the leak in miniature: `rn_image_picker_lib_temp_<uuid>.jpg` still carries a per-capture UUID, so the bare marker was the only option that leaks nothing on either platform. The count signal survives — one note per attachment.

The test now asserts one note per image and that neither URI string appears in the output, so restoring the URI-carrying note turns it red (verified by reverting). Not device-tested this round — the image path is still test-only, as flagged last round.

src/utils/__tests__/exportUtils.test.ts
sakebomb: Fixed in cfeb63c. You're right — the mock is return-only, and the argument observability the timestamp test relies on comes from `jest.fn()` recording calls, which was already true before this diff. Comment now says exactly that instead of implying the mock was made input-aware.

PR #857, opened 2026-08-19
chore(deps): upgrade llama.rn to 0.13.0-rc.0

a-ghorbani:
## Summary

Upgrade the `llama.rn` native dependency from `0.12.7` to `0.13.0-rc.0` (current npm `latest`), picking up the skipped `0.12.8` and `0.12.9` patch releases. Three files: `package.json`, `yarn.lock`, `ios/Podfile.lock`.

Upstream changes carried in:
- llama.cpp sync b10054 -> b10335 (via b10156, b10256)
- parallel-decoding state files valid and resumable for multimodal models (0.12.8)
- cancelled parallel completions now settle (0.12.9)
- TTS fixes: grammar enforcement, OuteTTS prompt/sampling, codec.cpp vocoder (experimental)

The pin is exact (no `^`) — a caret would resolve `<0.14.0` and could silently pull `0.13.0` final with different prebuilt binaries than anything verified here.

## Risk and verification

The one substantive risk was whether llama.rn still times the blocking completion path with its own wall clock, since the displayed generation rate depends on it and a pre-`0.12.7` regression made speculative turns read `0.00 tokens/sec`. Confirmed intact both statically and at runtime:

- `predicted_per_second = n_decoded / t_token_generation` in `cpp/rn-slot.cpp`, fed by `lm_ggml_time_us()` in `cpp/rn-completion.cpp`; `llama_perf_context` appears only as `_reset`, never as a rate source.
- Two speculative runs on the iOS simulator reported real rates with draft engagement: 48.39 tok/s and 43.84 tok/s. Never near zero.

Other checks:
- Both error-string couplings to native throws survive verbatim (`context is full`, `failed to create MTP draft context` — all three throw sites).
- The TTS breaking changes cannot reach the app: `initVocoder` / `getFormattedAudioCompletion` / `decodeAudioTokens` / `releaseVocoder` have zero call sites in `src/` (positive-controlled — the symbols are present in the typings and `llama.rn` imports do exist in `src/`). PocketPal's TTS is `rn-speech`/Supertonic.
- `n_parallel` stays pinned to 1 and there are no app call sites for llama.rn `loadSession`/`saveSession`, so the multimodal state-file fix is inert.
- Prebuilt slices intact: both iOS xcframework slices carry `rnllama.framework`; the Android `arm64-v8a` variant set is not reduced.

## Test plan

- `yarn lint` (0 errors), `yarn typecheck` (clean), `yarn test` — 270 suites / 4301 tests pass, 76.8% line coverage
- `pod install` in sync (`Podfile.lock` matches `Pods/Manifest.lock`), iOS Release build and Android release build both succeed
- iOS `quick-smoke` pass; Android `quick-smoke` pass
- iOS `speculative.spec.ts` 3/3 — engagement (draft tokens > 0 with a real generation rate), off-safety (non-MTP model resolves to off, no draft footer), crash-safety (width-mismatched draft does not abort the process)
- Cancellation: stop mid-generation settles, partial text retained, and the following turn completes — covering the 0.12.9 cancel fix

Visual evidence in a follow-up comment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ Versions are pinned in [`package.json`](package.json); the highlights:
 | UI | React Native Paper `5.14.5`, React Navigation |
 | State | MobX `6` (`mobx`, `mobx-react`, `mobx-persist-store`) |
 | Persistence | WatermelonDB (chat history), AsyncStorage (settings), Keychain (secrets) |
-| LLM | `llama.rn` `0.12.4` → llama.cpp · GGUF |
+| LLM | `llama.rn` `0.13.0-rc.0` → llama.cpp · GGUF |
 | TTS | `react-native-speech` `2.3.1` + `onnxruntime-react-native` `1.23.2` · ONNX |
 | Tooling | Yarn 1 (Classic), ESLint, Prettier, Jest, Husky + Commitlint |
 
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -59,7 +59,7 @@ PODS:
   - hermes-engine (0.82.1):
     - hermes-engine/Pre-built (= 0.82.1)
   - hermes-engine/Pre-built (0.82.1)
-  - llama-rn (0.12.7):
+  - llama-rn (0.13.0-rc.0):
     - boost
     - DoubleConversion
     - fast_float
@@ -3751,7 +3751,7 @@ SPEC CHECKSUMS:
   GTMAppAuth: 217a876b249c3c585a54fd6f73e6b58c4f5c4238
   GTMSessionFetcher: 5aea5ba6bd522a239e236100971f10cb71b96ab6
   hermes-engine: 273e30e7fb618279934b0b95ffab60ecedb7acf5
-  llama-rn: 98cb15ad698ad6264e4e9e7064500f8b66f88d05
+  llama-rn: e1b526654aa8c618ac793bf2bdf71447723f0386
   onnxruntime-c: 4a1a1a81da2087869dc9b6357f182952db2df29c
   onnxruntime-react-native: 4451eff6b1aa60589186c6a8422d436fba6a9192
   PromisesObjC: f5707f49cb48b9636751c5b2e7d227e43fba9f47
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "date-fns": "^4.1.0",
     "dayjs": "^1.11.18",
     "expr-eval": "^2.0.2",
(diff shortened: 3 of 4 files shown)


a-ghorbani:
## Visual evidence — llama.rn 0.13.0-rc.0 on iOS simulator








a-ghorbani:
## Review round 1 — follow-up

Merged `main` into the branch and re-verified everything on the merged tree.

### Why the merge mattered
The branch was cut before #835 landed, so every gate had run on a tree without it — and #835 rewrites the `ModelStore` path-recovery logic, the JS half of the same model-load path this PR swaps the native half of. The iOS build had also never compiled #835's `PalDataProvider.swift` against the new binaries. Both now verified together.

### Re-verified on the merged tree

| Gate | Result |
| --- | --- |
| TypeCheck | clean |
| Jest | 271 suites, 4313 passed (was 270 / 4301 — picks up #835's new tests) |
| Android build | BUILD SUCCESSFUL |
| iOS build | success, `PalDataProvider.swift` compiled against the new binaries |
| iOS `quick-smoke` | 1 passing |
| iOS `speculative` | **3 passing** — engagement, off-safety, crash-safety |
| Android `quick-smoke` | 1 passing |

Speculative engagement reports `22ms/token, 45.99 tokens/sec, 151ms TTFT` with `draft: 2947/3415 (86%)` — real draft engagement and a correctly-reported rate, which is the one behaviour this upgrade genuinely risked (the pre-0.12.7 timing bug read 0.00 tok/s on speculative turns).

### Install-size impact
Building a 0.12.7 control APK reproduced **247,390,011 bytes** against **265,483,290** for 0.13.0-rc.0 — a **+17.25 MB** delta on the two-ABI e2e APK, byte-reproducible, so it is the version rather than build variance. Consistent with a direct measurement of the `arm64-v8a` libraries (68.55 → 79.22 MB, +15.6%); iOS device binary grows 10.33 → 11.73 MB. Part of that is a new TTS/codec subsystem this app does not use.

### Also in this push
`README.md`'s stack table said `llama.rn 0.12.4`; it now tracks the real version.

### Known gaps, not addressed here
- `crash-safety` passes without exercising its property — it never selects a draft, so the width-mismatch comparison is unreachable. The SIGABRT path has no real runtime coverage.
- Paired-mode speculative was not run on this version.
- The cancellation check's "second turn completes" assertion resolves on the stopped turn's footer, so it cannot fail as written.

Each needs an e2e spec change, which does not belong in a dependency bump; they are recorded with concrete fixes for a follow-up.

PR #870, opened 2026-08-23
Per-row chat actions on Home: pin, rename, export, delete

a-ghorbani:
## Per-row chat actions on Home: pin, rename, export, delete

The bottom-tab migration unmounted `SidebarContent`, and with it the only call
site of `chatSessionStore.togglePinSession`. Pinned chat sessions shipped on
`main`, and after the nav swap the feature became unreachable on this branch —
the persisted `chat_sessions.pinned` column and the store's `Pinned` group had
no way to be set and no way to be seen.

Canonical Home already answers where it belongs. The populated frame
(`888:33822` "Homepage - default", **not** `888:33856` "first time user") ends
every chat row with a kebab (`888:33844` → `4359:1587`). We already drew that
kebab — as a bare `View` with no `onPress`. A control that looks tappable and
isn't is its own bug; it just never surfaced because the screen that was
captured for parity had no history rows.

### What changed

`HistoryRow`'s kebab is now a real control anchoring a per-row menu with the
same items the drawer offered — every one routing to a store method that already
existed:

| Item | Routes to |
| --- | --- |
| Pin / Unpin | `togglePinSession` |
| Rename | shared `RenameModal` → `updateSessionTitleBySessionId` |
| Export | `exportChatSession` (alerts on failure) |
| Delete | `deleteSession`, behind an `Alert` confirm; resets the active session first |

Pinned rows also **look and sort pinned** — filled star, ordered above unpinned.
A toggle whose effect is invisible isn't a feature; Home previously ordered by
recency alone and never read the flag.

### Scope notes

- **Markdown export was never unreachable.** I initially reported it lost along
  with pinning; it is not. `HeaderRight` is mounted (`ChatHeader` ← `ChatView`)
  and offers export-as-Markdown for the active session. Only pinning was
  genuinely orphaned.
- **Not re-homed:** bulk-selection mode (`Select…`, bulk delete/export). It was a
  drawer-list affordance with no canonical Home equivalent and nowhere to put a
  selection bar in a flat recency list. `bulkDeleteSessions` /
  `bulkExportSessions` stay unreferenced — they go with the `SidebarContent`
  deletion in the cleanup phase.
- **Designer ask (non-blocking):** canonical specifies the kebab but not the
  menu's contents or appearance — a metadata sweep of the Homepage section
  (`888:33821`) finds no dropdown frame. The item set is inherited from the
  drawer; presentation reuses the existing `Menu`/`Menu.Item`, the same
  DS-derived precedent as the chat-header overflow.
- **testIDs:** new controls use a `home-session-*` prefix, deliberately not
  `home-history-*`. An existing ordering test selects rows by
  `/^home-history-/`, and sharing the prefix would have silently widened it.

### Tests

Seven new tests, and they are gated rather than presence-asserting: reverting
the kebab to its old decorative `View` fails **6 of the 7**. They cover the
control's role/label, per-row menu isolation, each of the four writers, the
confirm-before-delete path, the Unpin label + star on a pinned row, and
pinned-first ordering.

- `yarn typecheck` — pass. `yarn lint` — pass (0 errors; 54 pre-existing warnings).
- `yarn test` — **281 suites / 4505 tests pass** (2 skipped).

### Docs

`app-shell.md` gains §4a (the row-action contract), three single-writer rows and
D13/D14; its deferred-cleanup entry that *gated* on re-homing pin is now
satisfied, so `SidebarContent` is dead in full and can be deleted outright in the
cleanup phase. `chat-flow.md` §8b is re-headed to record that `groupedSessions`
currently has no consumer while the pin writer is live again.

Visual evidence to follow on this PR.

---

### DIFF
--- a/src/locales/en.json
+++ b/src/locales/en.json
@@ -1054,7 +1054,12 @@
         "total": "Total: ",
         "usage": "Usage: "
       },
-      "byteSizes": ["Bytes", "KB", "MB", "GB"]
+      "byteSizes": [
+        "Bytes",
+        "KB",
+        "MB",
+        "GB"
+      ]
     },
     "chatView": {
       "menuItems": {
@@ -1650,6 +1655,7 @@
     "modelChipPrefix": "Model used",
     "modelChipEmpty": "Select a model",
     "chatHistory": "Chat history",
+    "sessionActions": "Chat actions",
     "emptyHint": "Select a pal or model, then start typing. Your conversations will appear here.",
     "sendLabel": "Send message",
     "searchLabel": "Search chats"
--- a/src/screens/HomeScreen/HomeScreen.tsx
+++ b/src/screens/HomeScreen/HomeScreen.tsx
@@ -1,5 +1,11 @@
-import React, {useContext, useEffect, useMemo, useState} from 'react';
-import {Image, ScrollView, Text, View} from 'react-native';
+import React, {
+  useCallback,
+  useContext,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
+import {Alert, Image, ScrollView, Text, View} from 'react-native';
 
 import {observer} from 'mobx-react';
 import {useNavigation} from '@react-navigation/native';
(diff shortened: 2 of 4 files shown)


a-ghorbani:
## Visual evidence — Home per-row chat actions (iOS sim, light + dark)

PR #873, opened 2026-08-23
Upgrade llama.rn to 0.13.0-rc.1 (llama.cpp b10588)

a-ghorbani:
## Description

Upgrades the vendored `llama.rn` native dependency from `0.13.0-rc.0` to `0.13.0-rc.1` — a pure llama.cpp sync of 253 builds, b10335 (`74ce157`) → b10588 (`70adb1b`).

The TypeScript surface of `llama.rn` is unchanged apart from `src/version.ts`, so no app code moves. `android/gradle.properties`, `android/build.gradle` and `rn-slot.cpp` are byte-identical between the two releases, so the from-source Android build mode and the native generation-rate path both carry forward unchanged. The one behavioural change in the package, an `embd.clear()` in `llama_rn_context_completion::embedding`, is unreachable here — PocketPal does not call the embedding API.

Also in this PR: the Android payload gate's symbol-count line now names the `llama.rn` version it read from `node_modules`, so `payload-report.txt` dates the Hexagon `.dynsym` count to the tree that was actually built rather than to whatever `package.json` declares. It is printed, never asserted — an unreadable `node_modules/llama.rn` prints `unknown` and leaves the verdict alone.

The Hexagon (NPU) backend evidence for this bump is the payload report from the `build-android` job — the SDK is provisioned there, so it is the build the project can guarantee produces the count. A build on a host without the SDK compiles the backend out and reports zero matches; that is compile health only, never a re-baseline.

## Platform Affected

- [x] iOS
- [x] Android

## Checklist

  - [x] iOS Simulator/Device
  - [x] Android Emulator/Device

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,7 +252,7 @@ Versions are pinned in [`package.json`](package.json); the highlights:
 | UI | React Native Paper `5.14.5`, React Navigation |
 | State | MobX `6` (`mobx`, `mobx-react`, `mobx-persist-store`) |
 | Persistence | WatermelonDB (chat history), AsyncStorage (settings), Keychain (secrets) |
-| LLM | `llama.rn` `0.13.0-rc.0` → llama.cpp · GGUF |
+| LLM | `llama.rn` `0.13.0-rc.1` → llama.cpp · GGUF |
 | TTS | `react-native-speech` `2.3.1` + `onnxruntime-react-native` `1.23.2` · ONNX |
 | Tooling | Yarn 1 (Classic), ESLint, Prettier, Jest, Husky + Commitlint |
 
--- a/ios/Podfile.lock
+++ b/ios/Podfile.lock
@@ -59,7 +59,7 @@ PODS:
   - hermes-engine (0.82.1):
     - hermes-engine/Pre-built (= 0.82.1)
   - hermes-engine/Pre-built (0.82.1)
-  - llama-rn (0.13.0-rc.0):
+  - llama-rn (0.13.0-rc.1):
     - boost
     - DoubleConversion
     - fast_float
@@ -3751,7 +3751,7 @@ SPEC CHECKSUMS:
   GTMAppAuth: 217a876b249c3c585a54fd6f73e6b58c4f5c4238
   GTMSessionFetcher: 5aea5ba6bd522a239e236100971f10cb71b96ab6
   hermes-engine: 273e30e7fb618279934b0b95ffab60ecedb7acf5
-  llama-rn: e1b526654aa8c618ac793bf2bdf71447723f0386
+  llama-rn: 46ed89f784eb5bc614657dcf31160f0d792675bc
   onnxruntime-c: 4a1a1a81da2087869dc9b6357f182952db2df29c
   onnxruntime-react-native: 4451eff6b1aa60589186c6a8422d436fba6a9192
   PromisesObjC: f5707f49cb48b9636751c5b2e7d227e43fba9f47
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "date-fns": "^4.1.0",
     "dayjs": "^1.11.18",
     "expr-eval": "^2.0.2",
-    "llama.rn": "0.13.0-rc.0",
+    "llama.rn": "0.13.0-rc.1",
     "marked": "^16.4.0",
     "mobx": "^6.15.0",
     "mobx-persist-store": "^1.1.8",
--- a/scripts/__tests__/verify-android-payload.test.js
+++ b/scripts/__tests__/verify-android-payload.test.js
@@ -352,6 +352,35 @@ describe('the Hexagon backend', () => {
     expect(status).toBe(0);
     expect(output).toContain('llama.rn unknown');
   });
+
+  it('reports the installed tree, not the declared pin, when the two disagree', () => {
+    const isolated = path.join(workspace, 'isolated');
+    fs.mkdirSync(isolated);
+    const copy = path.join(isolated, 'verify-android-payload.js');
+    fs.copyFileSync(SCRIPT_PATH, copy);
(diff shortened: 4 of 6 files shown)


a-ghorbani:
### Android payload gate — `build-android`, from-source with the Hexagon SDK

```
  arm64-v8a
    libraries: 12/12 present
    assets: 4/4 present
    extra rnllama libraries: none
    librnllama_v8_2_dotprod_i8mm_hexagon_opencl.so: 6899 .dynsym entries
      present  lm_ggml_backend_hexagon_reg
      present  lm_ggml_backend_is_hexagon
      16 .dynsym entries matching "hexagon" (declared 16), llama.rn 0.13.0-rc.1
  x86_64
    libraries: 4/4 present
    assets: 0/0 present

PASS: every declared library, asset and backend symbol is present.
```

The count is re-read from this build's `.dynsym` on every run, so an unchanged 16 is a fresh measurement at `0.13.0-rc.1`, not a value carried forward — and the count line now names the version it was read from. `scripts/android-payload-manifest.json` is deliberately absent from the diff: the symbol surface did not move across the sync, so there was nothing to re-declare.

No `opencl` / `codec_lm` string counts are offered as a control. The gate's own header records that `strings` false-positives on `codec_*_ht` and that two builds differing in whether Hexagon is compiled in have identical `opencl` counts, so those numbers cannot distinguish the two outcomes. The `.dynsym` read is the control.

### Other verification

| Check | Result |
| --- | --- |
| `yarn lint` / `yarn typecheck` | pass |
| `yarn test` | 274 suites, 4382 passed, 2 skipped |
| `pod install` (CocoaPods 1.15.2, `bundle exec`) | pass, `PODFILE CHECKSUM` unchanged |
| iOS Release simulator build | pass |
| Android `assembleE2eReleaseE2e` (local, darwin) | pass |
| `e2e/specs/features/speculative.spec.ts`, iOS simulator | 3/3 — off-safety, crash-safety, engagement |

Speculative decoding still engages after the 253-build llama.cpp sync: the engagement case observed `draft_tokens > 0` and a non-zero displayed generation rate, which is the assertion that would catch the `0.00 tok/s` mis-bucketing failure mode. No throughput figure is claimed from that run — simulator timings are not performance evidence, and draft-acceptance ratios swing run to run.

Lockfile churn is bounded: `yarn.lock` changes in one hunk inside the `llama.rn` block, and `ios/Podfile.lock` changes exactly four lines (the `llama-rn` version pair and its `SPEC CHECKSUMS` pair).

a-ghorbani:
## Independent review — APPROVE

Eight role subreviews (architecture, QA, security, performance, mobile, data, UX, local invariants) against a separate worktree built from the PR ref, followed by an adversarial refutation pass.

**0 blockers. 0 surviving concerns** — two concerns were raised and both were withdrawn under refutation; one was downgraded to a suggestion.

### Verification

All `NATIVE_CHANGES` obligations pass. CI green 4/4 on head.

The payload gate reports `16 .dynsym entries matching "hexagon" (declared 16), llama.rn 0.13.0-rc.1` with `lm_ggml_backend_hexagon_reg` and `lm_ggml_backend_is_hexagon` both present. This was re-derived directly from the shipped `.so` with `llvm-readelf` rather than read from the gate's own report, and the count is **16 at the base commit too** — no re-baseline slipped in. iOS is genuinely on the new sync (the b10588 commit string is present in the xcframework, sha256 matches the pin). 16 KB alignment clean on all 68 libraries. Lockfile churn is exactly one `yarn.lock` hunk and four `Podfile.lock` lines with `PODFILE CHECKSUM` untouched; the new SPEC CHECKSUM was independently reproduced.

Install size: **+3.44 MiB** total, **+2.49 MiB** on the arm64 slice — in line with the previous bump (+3.63 / +2.54), and essentially all compiled llama.cpp entering through pre-existing globs.

### Withdrawn on evidence

- **"DSP payloads have no content pin."** False premise — `bin/` is in llama.rn's published `files`, so the four DSP blobs ship inside the npm tarball, whose sha512 is pinned in `yarn.lock` (a line in this diff). Tarball digest and all five `.so` hashes verified against the installed tree.
- **"Native chat-parse layer changed and is untested."** Three of four evidence items were wrong or immaterial, and decisively, the proposed extra specs could not discriminate a regression: the relevant init function is byte-identical between versions and every candidate spec loads a model that never reaches the changed formats. They would pass regardless.
- **"Provenance attached to a conditional line."** Both mechanics reproduce, neither is a defect — per-artifact repetition is the report's structure, and the suggested fix would lose the token from the failure path, which is exactly where it earns its keep.

### Downgraded to suggestion

Both accelerator payloads rotated in this sync and neither executes anywhere in the verification set (CI has no NPU or Adreno; the iOS spec ran on a simulator). But every interface-defining header in `ggml-hexagon` is byte-identical, the host change is a fail-safe guard returning ops to CPU, the flash-attn changes are unreachable because Hexagon is pinned to flash-attn off, and backend-init failure degrades to CPU rather than failing hard. A Snapdragon smoke (pass/fail only) would still be cheap insurance.

### Follow-ups, none blocking

1. Harden the version read in the payload gate — validate the string and preserve the error class instead of collapsing every failure into `unknown`. A crafted version was shown to forge a report line; reachability is low since npm enforces semver and the lockfile digest was verified.
2. Prefer `require.resolve('llama.rn/package.json')` over a literal `node_modules` path, which silently yields `unknown` under hoisting.
3. One half of a new test assertion is unfalsifiable and should be renamed or documented as a forward guard.
4. Consider a per-ABI size ceiling: the arm64 slice is now ~154 MB and grows ~2.6 MB per sync.

### Worth a release note

`gguf.cpp` now type-checks `general.alignment` and returns cleanly instead of asserting — a malformed GGUF that **crashed the app** on the previous version now surfaces as a normal "failed to load".

PR #879, opened 2026-08-25
Strip unused CoreLocation symbols from the iOS binary (ITMS-90683)

a-ghorbani:
## Summary

Every App Store Connect upload returns the delivery warning **ITMS-90683**, asking for an
`NSLocationWhenInUseUsageDescription` purpose string. PocketPal uses no location at all. Rather than
declare a purpose string for an API the app never calls, this removes the CoreLocation surface from
the binary and adds a check that keeps it removed.

Two dependencies put it there:

- **`react-native-vision-camera` 4.7.2** — ten Swift files import CoreLocation, but the upstream
  `VISION_CAMERA_ENABLE_LOCATION` compilation condition covers only three of them.
  `LocationProvider`'s class body, the whole `CLLocationManager+requestAccess.swift` extension, and
  `MetadataProvider`'s four location sites are unguarded, so `$VCEnableLocation = false` alone removes
  call sites but not the API surface.
- **`react-native-device-info` 14.1.1** — `RNDeviceInfo.m` imports CoreLocation and exports
  `isLocationEnabled` / `getAvailableLocationProviders`. Their `RCT_EXPORT_METHOD` wrappers register at
  module-load time, so they cannot be dead-stripped. Neither JS API is called anywhere in the app.

## Changes

- `ios/Podfile` — `$VCEnableLocation = false`.
- `patches/react-native-vision-camera+4.7.2.patch` — extends upstream's own
  `#if VISION_CAMERA_ENABLE_LOCATION` to the three files that emit a CoreLocation symbol. The five
  value-type-only importers (`CLLocation`, `CLHeading`, `CLAuthorizationStatus`) are left untouched —
  they contribute no symbol. Upstreamable as a bug fix.
- `patches/react-native-device-info+14.1.1.patch` — removes the CoreLocation import and the two
  location method groups. That package offers no compile-time flag.
- `scripts/ios-api-surface-manifest.json` + `scripts/verify-ios-api-surface.js` — a check that reads a
  built `.app` or `.ipa` and asserts the declared surface is absent from **every** architecture slice,
  via `nm -u`, the `__TEXT,__objc_methname` section, and `otool -L`.
- `.github/workflows/ci.yml` — runs the check on the binary `build-ios` already builds, pinned with
  `-derivedDataPath derived-data`.

No `Info.plist` key is added, and no app behaviour changes. `<Camera>` never set `enableLocation`, so
the GPS EXIF and QuickTime-location branches were already unreachable.

## How the check avoids passing on nothing

An empty grep proves nothing about a binary that may not have been read at all, so:

- **Sentinels.** The manifest names references the app's shipped camera guarantees
  (`_OBJC_CLASS_$_AVCaptureDevice`, `requestAccessForMediaType:completionHandler:`). Every reading that
  asserts an absence asserts these present, in the same invocation over the same bytes. A missing
  sentinel is an instrument failure, never a clean result.
- **It fails closed** on an unlocatable binary, a tool that will not run, an empty symbol table, an
  empty selector section, or zero slices.
- **Manifest floors.** Emptying a list is the cheapest edit that unblocks a build, so `loadManifest`
  refuses a manifest with no forbidden surface, with `symbolPatterns` or `selectors` cleared (both, not
  either — they read different surfaces), with patterns that no longer contain the framework's own
  name, or with an exemption or sentinel carrying no reason.
- **Broad patterns, named exemptions.** An exact forbidden list cannot fail on a symbol nobody
  predicted. `__swift_FORCE_LOAD_$_swiftCoreLocation` is the single exemption — an autolink hint
  emitted by any file importing the overlay, including the value-type importers that must keep
  compiling. A symbol the pattern selected by accident is an over-match, and the repair is to correct
  the pattern, never to exempt the symbol.
- **69 tests**, split so both halves are actually exercised: the verdict logic runs on any platform
  against deliberately broken in-memory readings, and a macOS-only suite compiles real fat Mach-O
  fixtures — clean, location-using, and mixed-slice — and drives the check end to end through both
  `--app` and `--ipa`. A rule seen only passing is not known to be able to fail.

## Evidence

The check was written and run against a binary **before** the fix, so it was seen failing on real bytes
first. Both readings are Release simulator builds of this app, same check, same manifest:

| | symbols matching the patterns | forbidden selectors | `otool -L` | exit |
| --- | --- | --- | --- | --- |
| before | `_OBJC_CLASS_$_CLLocationManager`, `_kCLLocationAccuracyNearestTenMeters` (+ the exempted autolink hint), both slices | 7/7, both slices | `CoreLocation.framework`, `libswiftCoreLocation.dylib` | **1** |
| after | only the exempted autolink hint, both slices | 0/7, both slices | `libswiftCoreLocation.dylib` | **0** |

Sentinels present in every slice of both readings. `CoreLocation.framework` also dropped out of
`LC_LOAD_DYLIB` entirely, which the design had not assumed it would.

## Verification

`yarn lint` (0 errors), `yarn typecheck`, and `yarn test --coverage` (276 suites, 4452 passed,
76.9% statements / 71.7% branches, threshold 60%) all pass.

`pod install` is clean and logs `[VisionCamera] $VCEnableLocation is set to false!`;
`ios/Podfile.lock` moves only VisionCamera's checksum.

**Android** is verified as a `prodDebug` APK (`yarn build:android`), not a release build — a local
release build risks printing the signing password, so the release variant is left to this PR's own
`build-android` job. What the debug build does prove is the risk the shared `react-native-device-info`
patch introduces: it compiles for Android. The merged manifest carries 15 permissions and no location
permission of any kind (`android.permission.CAMERA` present as a positive control).

## Scope, stated rather than implied

- **The check covers CI, not the release path.** `release_ios_testflight` builds and uploads in one
  lane, so a gate there would run after TestFlight already held the IPA; splitting that lane is left to
  a follow-up. The release artifact is therefore covered only transitively, through the ref CI ran on.
- **The subject is the simulator binary, not the shipped one.** Sound for these two dependencies
  because neither guards its CoreLocation use behind `#if targetEnvironment(simulator)`. A dependency
  that did would not be caught today.
- **Nothing this PR examines is ever uploaded to App Store Connect**, and ITMS-90683 is emitted against
  the uploaded IPA. The warning's clearance is confirmed by the next TestFlight delivery and by nothing
  before it. If it persists on a build carrying both patches, framework linkage with no symbol
  reference is the next hypothesis.
- **`patches/` now pins third-party source in four places.** A dependency bump that moves a patched
  file is caught by this check on the bump PR, whether the patch fails to apply or applies and no
  longer suffices. It is not caught by `yarn install`: `scripts/postinstall.sh` runs `npx patch-package`
  with no `set -e` and not as its last command, so a failed patch still exits 0. Worth a follow-up
  independently of this change, since every patch rests on it.


a-ghorbani:
Closing in favour of a much smaller fix.

This branch removed the CoreLocation symbols by patching two dependencies (`react-native-vision-camera`, `react-native-device-info`) and added a CI gate to keep them out. It worked — the symbols are gone and the binary is 596 KB smaller — but the cost is permanent: two `patch-package` patches pinned to exact versions, to be re-created against restructured upstream source on every bump of either dependency, plus an instrument to guard them.

ITMS-90683 is a non-blocking delivery warning. Declaring the purpose string Apple asks for is one line, works permanently, and survives every dependency bump. That is the trade worth making here.

Superseded by the Info.plist change.

PR #880, opened 2026-08-25
Settings account surface: log in, create account, account settings

a-ghorbani:
## Settings-side account surface: log in, create account, account settings

`SettingsScreen` hardcoded `const IS_REGISTERED = false`. Shipping consequences on this branch:
the "Create Account" CTA rendered **`disabled`**, the Account Settings row was **`inert`**, the
registered header interpolated **empty strings**, Log out had **no `onPress` at all**, and the file
never imported `authService` — so a signed-in user still saw "Create Account". Auth worked, but only
from one place: `ExploreScreen` mounting `AuthSheet`.

This adds three pushed routes — **Log in**, **Create Account**, **Account Settings** — plus a reset
sheet, and wires the launcher to real session state.

### Scope: the drawn flow needs backend that does not exist

Canonical entry (`3011:26556`) is **email-first**: one email field → `Get Started` → routes to
"Welcome Back" (existing account) or "Create Account" (new). That fork requires knowing whether an
email already has an account, and **Supabase deliberately does not expose that** — it is a
user-enumeration vector. Not a missing wrapper; a capability the backend declines to offer.

So this slice ships **separate Log in and Create Account entry points** on the flow the backend
supports, in the design's visual language. Deliberately out, each with no backing method:
in-app set-new-password (`resetPassword` redirects to the **web**; `detectSessionInUrl:false` and
`DeepLinkStore` has no auth handling), verification resend, **Sign in with Apple** (drawn on all
three entry screens), delete account, change password. Both Account Settings screens are
substantially those last two, so this one lands thin by design — identity line + username only.

Every omission and every addition is registered, not absorbed. `AuthService` is **not touched**
(zero diff); no auth method is added.

### What is built but NOT drawn — needs a designer pass

Splitting the router forces four things no render shows: an **email field on Log in**
(`3011:26630` draws password only), **Log in ⇄ Create account cross-links**, a **legal footer on
Log in**, and the **`settings-log-in`** link on the launcher card. Log in also ships **no subtitle** —
the drawn one ("Enter the password used for the account sam@gmail.com") stops making sense once the
email is typed on that same screen, and shipping placeholder copy would enter the 13-language
Weblate queue for a string that may never exist.

### Contract decisions worth review

- **`saved-unconfirmed` is a distinct outcome, not an error.** `updateProfile` returns `void` and its
  read-back fails silently three ways — Supabase error, throw, and `PGRST116` nulling `profile`. So a
  write can land while the confirming read does not. Reporting that as a failure would be untrue.
- **Errors are snapshotted, never live-bound.** `onAuthStateChange` nulls `authService.error` on
  *every* auth event including `TOKEN_REFRESHED`, so a live binding can blank a displayed error with
  no user action.
- **Screen-local `submitting`**, never `authService.isLoading` — that flag is singleton-wide, so a
  checkout sign-in would move a Settings spinner.
- **Submit stays pressable with empty fields.** Disabling it made all three `validation.*Required`
  strings unreachable. The failure is explained, not hidden.
- **Google's outcome is observed, not returned.** `signInWithGoogle` returns `void`; post-await
  `isAuthenticated` is explicitly *not* a success test.

### Five defects the capture pass caught that the test suite could not

Each was found by rendering on an iPhone 17 Pro sim, and each is invisible to Jest by construction:

1. **Submit invisible in dark, empty state.** The DS Button's disabled fill is 8 % opacity — invisible
   on black. Repainted at the call site (`surfaceVariant`); **no DS file touched** (theming I_UI5).
2. **RTL: field labels did not mirror.** Cause was the DS `Input`'s internal `label`, which sets no
   `textAlign` and fell back to first-strong. Moved to a call-site label.
3. **RTL: prompt rows read backwards** — *"Create account Don't have an account?"*. Two siblings in a
   mirrored row; now one text run.
4. **Inline links lost their `testID` natively** — a regression from fixing (3). Nested RN `Text`
   becomes attributed-string ranges in one `RCTTextView`; RN synthesizes children carrying label and
   role but **never** `testID`. Fixed by moving identity to the parent `Text`, then **verified in the
   live XCUITest tree**.
5. **Onboarding back chevron near-invisible in dark.** Not from this slice — the exported SVG baked
   `fill="#181715"`. Note a call-site `fill=` would *not* have helped: `.svgrrc` rewrites only
   `#333333`/`#858585`/`#FF653F` into `{props.fill}`, so any other hex stays literal and the path stops
   inheriting from the root `<Svg>`. Now `currentColor` + an explicit `color`. **Review this commit
   separately — it is onboarding code, already on `main` via an earlier slice.**

### Tests

**285 suites / 4554 passing**, typecheck and lint clean. Scenario coverage maps 1:1 to the design
contract's canonical scenarios A–I.

Two tests deserve calling out because they cover *different* properties and neither sees the native
layer: one proves the RTL phrase composes as a single text run (it **stayed green throughout defect 4**),
the other proves the structural precondition for RN to emit an identifier. Both were positive-controlled
by restoring the broken shape. The `title: ''` assertion was likewise positive-controlled — removing the
option makes it fail with `Expected: "" / Received: undefined`.

### Evidence

29 captures in the story dir, light + dark + forced-RTL, across three passes (v1 → v3 as defects were
fixed). RTL revert verified **byte-identical** to the pre-RTL captures.

**Five states could not be captured.** The configured Supabase host (`192.168.0.92:54321`) is
unreachable — `curl` → `000`, ping 100 % loss, re-checked across three separate passes, while the sim's
own internet works. Measured consequence: sign-in errors after ~76 s with `Network request timed out`.
Google sign-in launches correctly through to Google's consent page but exchanges its token against the
same dead host. Blocked: `settings-registered`, `account-details` (both modes), the RTL pass over
account-details, and the verification state. Nothing was faked or silently skipped — documented-failure
path per `docs/workflows/visual-capture.md`. **The registered-state screens in this PR have never been
seen rendering.**

### Docs

`context/architecture/settings.md` is authored and deliberately **uncommitted** — the human lands
`context/architecture/*` per-slice. It carries the new IA, the writer table, invariants, the decision
register, and a deviation register split into drawn-not-built / built-not-drawn / cross-doc traps.

---


a-ghorbani:
## Visual evidence — account screens (iOS sim; light, dark, forced-RTL)

PR #890, opened 2026-09-02
i18n(zh): translate sidebar menu item 'Pals' to '伙伴'

Hmission:
## Summary
zh.json `sidebarContent.menuItems.pals` remains untranslated ("Pals") while all sibling menu items (聊天 / 模型 / 基准测试 / 设置 / 关于) are localized. This aligns with the existing `menuItems.pals → "Pals（实验性）"` translation at the settings submenu.

## Motivation
- Single missing key in the Chinese locale (verified `l10n:validate`-style scan: all other sibling keys are translated)
- Pocket Chick (fork) carries the same translation; upstreaming keeps source of truth in one place

## Change
- `src/locales/zh.json`: `"pals": "Pals"` → `"pals": "伙伴"`

No code change, JSON only.

### DIFF
--- a/src/locales/zh.json
+++ b/src/locales/zh.json
@@ -955,7 +955,7 @@
       "menuItems": {
         "chat": "聊天",
         "models": "模型",
-        "pals": "Pals",
+        "pals": "伙伴",
         "benchmark": "基准测试",
         "settings": "设置",
         "appInfo": "关于",


a-ghorbani:
Thanks for the PR, and for taking the time to look at the Chinese locale.

Two reasons I'm going to close this one:

**1. Locale files aren't edited in the repo.** Per [CONTRIBUTING.md](https://github.com/a-ghorbani/pocketpal-ai/blob/main/CONTRIBUTING.md#translations-localization), only `src/locales/en.json` is edited here — every other locale is managed by translators through [Weblate](https://hosted.weblate.org/translate/pocketpal-ai/translations/zh_Hans/) and is overwritten on the next sync. So even if merged, this change would be reverted automatically.

**2. The string isn't actually untranslated.** In Weblate the `components.sidebarContent.menuItems.pals` unit for `zh_Hans` is marked **Translated**, with `Pals` as the chosen target — a deliberate decision, not a gap. "Pal" is a product term in PocketPal, and the zh locale keeps it in Latin consistently: 51 strings do so (我的 Pal (本地), 发现 Pal, 导出所有 Pal, 搜索所有 Pal…). The `screenTitles.pals` → `Pals（实验性）` entry cited as precedent actually keeps *Pals* and localizes only the "(experimental)" qualifier. And 伙伴 already carries the generic "companion" sense in the zh onboarding copy (小伙伴), so localizing the product term to it would blur the two.

Whether the Pal term *should* be localized in Chinese is a fair question, and I'm genuinely undecided. But it's a decision for all 51 occurrences at once, not one key — so the right place for it is [Weblate](https://hosted.weblate.org/translate/pocketpal-ai/translations/zh_Hans/) (comment on the unit, or open a discussion) rather than a JSON PR. Contributions there are very welcome.

PR #894, opened 2026-09-04
Count cache-reused prompt tokens in the remote context total

a-ghorbani:
## What this fixes

A remote turn's used-token total dropped the part of the prompt the server served from
its KV cache, so the context meter under-reported — badly, and worst exactly when a
conversation was long enough for the meter to matter.

Measured against a live `llama-server` (build `b9976-e3546c794`): a repeated prompt
reported `prompt_n = 1` and `cache_n = 17`, a true prompt of 18. **The banner saw 1.**

`completionTypes.ts` already documented this as a known limitation — *"tokens_cached is
not exposed at the engine boundary, so on prompt-cache-reuse turns it under-counts KV
occupancy"*. That note was wrong twice over, which is why the fix is small:

- **The local path never had the bug.** `JSICompletion.h:247` sets `tokens_evaluated`
  from `slot->num_prompt_tokens`, which `rn-slot.cpp:172` sets to `tokens.size()` — the
  whole prompt.
- **The value was never unexposed.** llama.rn declares both `tokens_cached` and
  `timings.cache_n`; this app's own mapping dropped them.

So the missing term was already arriving on every finish chunk. No new request, no new
store field, no new failure branch.

## Also fixed: the remote context-full banner never rendered

Three sites wrote `used: 0` alongside `contextFull: true`, reading an `n_ctx` that is
never set for a remote model. The resolver's freshness gate requires
`used >= effectiveNCtx - AUTOCLEAR_RUNWAY`, which `0` always fails — so the banner was
unreachable for every remote session. Those sites now read the same resolved window the
banner itself reads, and write *absent* rather than `0` when the window is unknown.

A fourth site was a parameter that was never read — the trap that produced the other
three.

## The identity this rests on

```
input_tokens  ==  usage.prompt_tokens  ==  timings.prompt_n + timings.cache_n
```

Verified 5/5 on a live server across cold, cache-reuse, tool-carrying, thinking-disabled
and long-prompt turns, then re-verified on the **streaming** path the app actually uses:

| case | `input_tokens` | `prompt_n` + `cache_n` | |
| --- | --- | --- | --- |
| cold | 32 | 32 + 0 | match |
| **cache reuse** | **32** | **1 + 31** | **match — the load-bearing case** |
| with a tool schema | 161 | 161 + 0 | match |
| `enable_thinking: false` | 34 | 28 + 6 | match |
| long prompt | 314 | 311 + 3 | match |

The cache-reuse row carries the weight: it is the only case where the two sources are
computed differently enough to drift, and it is the case this fix exists for.

Upstream asserts the arithmetic itself (`test_completion.py:659`,
`test_kv_keep_only_active.py:80`) — but **on the native `POST /completion`, read
non-streaming from `res.body`**. This app reads the OpenAI-compatible endpoint's SSE
finish chunk. Endpoint *and* delivery differ, so those tests constrain the field names
and their relationship, not our path; the streaming observation is ours.

## `undefined` is not `0`

`timings.cache_n` landed in **b6399 (2025-09-06)**. An older build emits no key at all,
and that is a different fact from a cold prompt reporting `0` — which is common and
real. Absent degrades to the previous value, never below it; a reported `0` is a count.
Both are covered by tests that catch *different defects* rather than different values:
one fails a naive sum (`NaN`), the other fails a shape that reads a real `0` as absent.

## Evidence and its limits

Every wire fact here was measured on **`b9976-e3546c794`, 2026-07-11, Linux aarch64**.
Cross-build generalisation is unverified. The fixture is a verbatim nine-key capture;
tests project their inputs from it and keep expectations as independent literals, so an
edited fixture fails loudly rather than moving the goalposts.

Three claims about upstream coverage were checked and withdrawn during this work:

- Upstream does **not** assert `input_tokens == prompt_n + cache_n`. Every assertion on
  that endpoint is a loose bound (`> 5`, `> 10`) or relational — all would still pass if
  the count drifted from what the completion charges.
- The timings assertions run against the **native** endpoint, non-streaming (above).
- `test_compat_anthropic.py:882` exercises `POST /v1/messages/count_tokens`, a
  **different endpoint**. It shares the `input_tokens` field *name*, which is what made
  the citation look right.

## Scope

The pre-send token count and the persistent gauge are **not** in this change. They moved
to a follow-up, together, because the count has no home without a surface and the surface
is a larger change than it appears: `effectiveContextLength` is defined for local models
too, so a permanent gauge is either arbitrarily remote-only or a chat-wide change
touching a read path this PR deliberately leaves alone — plus copy and an a11y label
across 19 locales.

Splitting that way also puts the wider-reaching half first: `cache_n` predates the count
endpoint by nine months and ~3,600 builds, so this fix works on essentially every
`llama-server` in the wild.

## Architecture

The remote-servers and chat-flow architecture docs are updated alongside this change: the
token-accounting rule now reads `prompt_n + cache_n`, and a new section records where the
context window comes from and what each `used` state means.

---

### DIFF
--- a/jest/fixtures/llamaServerTimings.ts
+++ b/jest/fixtures/llamaServerTimings.ts
@@ -9,22 +9,12 @@
  * aarch64). Every wire fact here was observed on that one build; cross-build
  * generalisation is unverified.
  *
- * Three things this capture must not be read as saying. Each is a citation that
- * looks right, and each was written down before being checked against the
- * source lines:
- *
- *  - Upstream's own `timings` assertions (`test_completion.py:659`,
- *    `test_kv_keep_only_active.py:80`) run against the native
- *    `POST /completion` and read the response body non-streaming, while this
- *    app reads the OpenAI-compatible endpoint's SSE finish chunk — a different
- *    endpoint and a different envelope. Upstream does not cover this path, and
- *    this fixture is the app's own capture rather than a copy of theirs.
- *  - Upstream asserts nothing about the join between a separately counted
- *    prompt and `prompt_n + cache_n`. Its token-count endpoint has exactly two
- *    tests and both are loose bounds (`> 5`, `> 10`).
- *  - `test_compat_anthropic.py:882` is a different endpoint,
- *    `POST /v1/messages/count_tokens`, which merely shares the `input_tokens`
- *    field name. It constrains nothing here and must not be cited for it.
+ * Upstream's own `timings` assertions (`test_completion.py:659`,
+ * `test_kv_keep_only_active.py:80`) run against the native `POST /completion`
+ * and read the response body non-streaming, while this app reads the
+ * OpenAI-compatible endpoint's SSE finish chunk — a different endpoint and a
+ * different envelope. Upstream does not cover this path, and this capture is
+ * the app's own rather than a copy of theirs.
  */
 export const cacheReuseTimings = {
   cache_n: 31,
--- a/jest/fixtures/llamaServerTimings.ts
+++ b/jest/fixtures/llamaServerTimings.ts
@@ -1,0 +1,39 @@
+/**
+ * Verbatim `timings` object from a llama-server SSE finish chunk, captured on a
+ * cache-reuse turn: `prompt_n` counts only the newly evaluated tokens and
+ * `cache_n` the reused prefix, so the prompt total is their sum. All nine keys
+ * are as they arrived — a stand-in trimmed to the fields a reader finds
+ * interesting would test the parser against a shape no server emits.
+ *
+ * Captured from llama-server build `b9976-e3546c794` (2026-07-11, Linux
+ * aarch64). Every wire fact here was observed on that one build; cross-build
+ * generalisation is unverified.
+ *
+ * Three things this capture must not be read as saying. Each is a citation that
+ * looks right, and each was written down before being checked against the
+ * source lines:
+ *
+ *  - Upstream's own `timings` assertions (`test_completion.py:659`,
+ *    `test_kv_keep_only_active.py:80`) run against the native
+ *    `POST /completion` and read the response body non-streaming, while this
+ *    app reads the OpenAI-compatible endpoint's SSE finish chunk — a different
+ *    endpoint and a different envelope. Upstream does not cover this path, and
+ *    this fixture is the app's own capture rather than a copy of theirs.
+ *  - Upstream asserts nothing about the join between a separately counted
+ *    prompt and `prompt_n + cache_n`. Its token-count endpoint has exactly two
+ *    tests and both are loose bounds (`> 5`, `> 10`).
+ *  - `test_compat_anthropic.py:882` is a different endpoint,
+ *    `POST /v1/messages/count_tokens`, which merely shares the `input_tokens`
+ *    field name. It constrains nothing here and must not be cited for it.
+ */
+export const cacheReuseTimings = {
+  cache_n: 31,
+  prompt_n: 1,
+  prompt_ms: 6.454,
+  prompt_per_token_ms: 6.454,
+  prompt_per_second: 154.94267121165169,
+  predicted_n: 3,
+  predicted_ms: 12.9,
+  predicted_per_token_ms: 4.3,
+  predicted_per_second: 232.5581395348837,
+};
--- a/src/api/__tests__/openai.test.ts
+++ b/src/api/__tests__/openai.test.ts
@@ -1610,6 +1610,32 @@ describe('streamChatCompletion', () => {
     expect(result.timings?.cache_n).toBeUndefined();
   });
 
+  it('guards each timings token key independently (only cache_n)', async () => {
+    const resultPromise = streamChatCompletion(
+      {messages: [{role: 'user', content: 'Hi'}], model: 'test-model'},
+      'http://localhost:1234',
+    );
+
+    const xhr = MockXHR.instances[0];
+    xhr.simulateHeaders(200);
+    xhr.simulateProgress(
+      'data: {"choices":[{"delta":{"content":"Hello"},"finish_reason":null}]}\n\n',
+    );
+    xhr.simulateProgress(
+      `data: {"choices":[{"delta":{},"finish_reason":"stop"}],"timings":${JSON.stringify(
+        {cache_n: cacheReuseTimings.cache_n},
+      )}}\n\n`,
+    );
+    xhr.simulateProgress('data: [DONE]\n\n');
+    xhr.simulateLoad();
+
+    const result = await resultPromise;
+    // A fully reused prompt evaluates nothing, so the whole count arrives on
+    // cache_n and prompt_n is absent. Reading prompt_n alone loses it.
+    expect(result.tokens_evaluated).toBe(31);
+    expect(result.timings?.prompt_n).toBeUndefined();
+  });
+
   it('returns no timings when server does not provide them', async () => {
     const resultPromise = streamChatCompletion(
       {messages: [{role: 'user', content: 'Hi'}], model: 'test-model'},
--- a/src/api/__tests__/openai.test.ts
+++ b/src/api/__tests__/openai.test.ts
@@ -13,6 +13,7 @@ import {
   directVisionModelsBody,
   routerModelsBody,
 } from '../../../jest/fixtures/remoteModelList';
+import {cacheReuseTimings} from '../../../jest/fixtures/llamaServerTimings';
 
 /** Build a minimal Headers-like object for fetch mocks. */
 function mockHeaders(entries: Record<string, string> = {}) {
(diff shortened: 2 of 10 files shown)


a-ghorbani:
## ⚠️ Visual evidence captured but NOT YET POSTED — pending a token

`post-pr-visual-evidence.sh` returned `MANUAL_POST_REQUIRED`: no GitHub session token is
available for image upload. **The captures exist and are verified, but they are not on
this PR, so the visual-evidence gate is NOT satisfied.** This comment records the pending
condition; it is not a substitute for the images.

### What was captured

A pre/post pair on a **real** `llama.cpp` server (`b9976-e3546c794`, two models already
loaded) driven on a physical device — not a mock, because the entire claim of this change
is that a number is correct.

| turn | before (`main`) | after (this branch) |
| --- | --- | --- |
| **1 — cold prompt, no cache reuse** *(control)* | context-warning, **97%** | context-warning, **97%** — unchanged |
| **2 — cache-reusing turn, truly 8192/8192** | **no banner at all** | **context-full, 100%** |

Turn 1 is a deliberate positive control. A cold turn has no reused prefix, so the added
term cannot move it — and it does not. That is what makes the turn-2 difference
attributable to the missing `cache_n` term specifically, rather than to a banner that was
simply broken before.

Turn 2 on `main` renders nothing because the under-count reads 523 tokens (6%), below the
80% threshold, while the server's own accounting for that turn is
`prompt_n 41 + cache_n 7669 + predicted 482` — a full window. Both `02` captures also
carry in-image proof of prefix reuse: time-to-first-token collapses from ~1.3–1.5 s on the
cold turn to ~150 ms on the reusing turn.

### Files, pending upload

```
e2e/debug-output/screenshots/visual-captures/TASK-20260903-1954/pre/01-cold-turn-control.png    md5 afb3f043…
e2e/debug-output/screenshots/visual-captures/TASK-20260903-1954/pre/02-cache-reuse-turn.png     md5 3a976ca4…
e2e/debug-output/screenshots/visual-captures/TASK-20260903-1954/post/01-cold-turn-control.png   md5 b625a8b2…
e2e/debug-output/screenshots/visual-captures/TASK-20260903-1954/post/02-cache-reuse-turn.png    md5 52543d68…
```

These paths are gitignored and local to the machine that produced them.

---

a-ghorbani:
## Independent review

**Reviewers**: code, QA, architecture, local-invariants, UX — five lenses, applied independently.
**Verdict**: **REQUEST_CHANGES** — all five lenses concur.

`counts: { blocker: 1, concern: 3, suggestion: 8 }`

---

## Verification gate

| check | result |
| --- | --- |
| eslint (10 changed files) | PASS (0) |
| `tsc --noEmit` | PASS (0) |
| test suite | PASS — 274 suites / 4395 passed / 2 skipped |
| changed-surface coverage vs 60% floor | PASS — lowest axis `BannerRow` branches 60.0, `useChatSession` branches 61.7 |
| `NATIVE_CHANGES=NO` on the commit range | PASS — 0 `android/` `ios/` `package.json` gradle paths (three-dot) |
| **visual evidence posted** | was `NOT_RUN` at review time · **now posted** |

`APPROVE` is mechanically invalid while any required verification is `NOT_RUN`.

---

## BLOCKER — since resolved

**B1 — visual evidence was not posted at the time of this review.** The trigger was real:
this change adds a percent readout beside the full banner's copy, which is a sentence in
four variants, so how the numeral sits against wrapped multi-line text is something a
capture settles and a description cannot.

**Resolved** — the before/after captures are now posted on this PR.

## CONCERNS

### C1 — The "unknown count" guard keys on the envelope, not the count

`src/hooks/useChatSession.ts:248-253`

```ts
const used = isRemote && !result.timings
  ? undefined
  : (result.tokens_evaluated ?? 0) + (result.tokens_predicted ?? 0);
```

**Found independently by four reviewers.** `openai.ts` leaves `tokens_evaluated` undefined
whenever `timings` is present but carries neither `prompt_n` nor `cache_n`. On that shape
`?? 0` fires and `used` becomes the **predicted-only tally** — precisely what the comment
three lines above, `chat-flow.md` the remote context-accounting section all forbid.

Proven at runtime, not argued: a finished remote turn with
`{timings: {predicted_per_second: 80, prompt_per_second: 800}, tokens_predicted: 600}`
yields `snapshot.used = 600`.

**Refutation partly succeeds**, which is why this is CONCERN and not BLOCKER: a real
llama-server finish chunk always carries `prompt_n`, the value is inert today (it fails
both the freshness gate and the 0.8 threshold, so no banner renders), and the probe is
`serverType === 'llama.cpp'`-gated. It stops being inert in the successor item, which
renders `used` whenever it is defined. The repo's own untouched
`openai.test.ts:1563` already pins the token-less shape as real llama.cpp behaviour.

**Fix**: gate on `result.tokens_evaluated === undefined`. That also makes the `isRemote`
branch unnecessary and closes the same hole on the local path.

**The fix needs a test with it.** Applying it leaves the entire suite passing — nothing
currently pins either the present or the corrected semantics.

### C2 — `chat-flow.md` contradicts itself about the model this PR ships

`context/architecture/chat-flow.md:89-96` still states the remote counts are "sourced from
server `timings.prompt_n` / `predicted_n`" and says nothing about `used` being optional,
while the remote context-accounting section — same file, ~1300 lines later — states the corrected model. A reader who finds
the first passage gets the pre-PR contract.

Two smaller instances in the delta this lane wrote: the remote context-accounting section says "the **three** catch paths" in
one paragraph and "the **two**" in the next (three is right), and claims all four writers
"resolve the window through that one selector" — false for `deriveSnapshotFromResult`,
whose window parameter **this PR deleted**. the remote context-accounting section's state table also omits a fifth reachable
state: the catch paths assign the window itself as a synthetic count, which with the new
readout renders **100%**.

Doc-only, dev-team repo, no code change.

### C3 — The fixture header is case-1 comment over a case-4 core

`jest/fixtures/llamaServerTimings.ts:1-28` — a 28-line header on a 10-line object. The
provenance paragraph is genuine case 4 and must stay. The meta-sentence ("each was written
down before being checked against the source lines") records **how the author reached the
answer**, and two of the three refutation bullets rebut citations that appear nowhere in
the app repo, with no repo, path or pinned commit — dangling as written.

**Fixture integrity itself passes**: all nine keys present, and the values are
arithmetically self-consistent (`1000/6.454 = 154.9427`, `12.9/3 = 4.3`,
`1000/4.3 = 232.5581`) — a real capture, not values edited to suit a test.

---

## A disagreement between reviewers, resolved by the synthesizer

Local-invariants reported that the `cache_n` 0-vs-absent test "passes against the code it
exists to pin" — reverting `openai.ts` to `origin/main`'s `prompt_n`-only form left it
green. QA reported the opposite: the trap is closed, with three mutations killed.

**Both observations are correct; the finding is refuted.** Two mutations, run here:

| mutation | `distinguishes a reported cache_n of 0 from an absent one` |
| --- | --- |
| drop the `cache_n` term from the sum (`origin/main`'s form) | **passes** |
| `cache_n ? sum : undefined` — read a real `0` as absent | **fails ✕** |

The test pins **its own claim** — that a reported `0` is a count and an absent key is not —
and is killed by the mutation that breaks that distinction. The mutation local-invariants
applied breaks the **sum**, which is a different test's responsibility and is caught by
`adds the cache-reused prefix to the evaluated prompt count` (verified failing under that
mutation). Reading one test's pass under another test's mutation as a defect is a
mis-scoped check, not a gap.

**What survives from that thread is narrower and real** (recorded as a suggestion below):
guarding on `prompt_n` alone survives the suite, because there is a `(only prompt_n)` case
and a `(only predicted_n)` case but no `(only cache_n)` case. Half the "guarded separately"
claim is comment-only.

---

## SUGGESTIONS

1. No `(only cache_n)` test — guarding on `prompt_n` alone survives all 177 tests.
2. `expect('used' in snapshot).toBe(true)` pins a key that `JSON.stringify` drops on the way
   to the repository, so the durable snapshot has no such key.
3. The third rewritten `used` writer (`useChatSession.ts:884-889`) is uncovered by any test.
4. The full banner can read "ran out of room" beside **88%**: the freshness gate opens a
   flat 256 tokens early, so at the default 2048 window the variant is reachable from 1792.
   Wider for remote, where `effectiveContextLength` is the **per-slot** window. Either
   render 100% on the full variant or keep the numeral on the warning only.
5. The percent is an unlabelled a11y orphan — announced once as a bare "97 percent" with no
   noun. The meter is correctly hidden, so this is the only accessible fullness signal.
   Costs one l10n key.
6. `BannerRow.tsx:134` still does `Math.round((ratio ?? 0) * 100)`, coercing unknown to
   zero — the thing `completionTypes.ts:81` now forbids. Unreachable today.
7. The `prompt_n`/`cache_n` rationale is restated in prose at five sites; keep it at the
   mapping site.
8. `ratio != null` on the full variant is dead — the resolver only returns that variant
   from inside the branch where `ratio` is computed.

---

## What was verified and holds

- The fix's premise is independently sound: llama.rn sets `tokens_evaluated =
  num_prompt_tokens`, the **whole** prompt, so this change *unifies* the field's meaning
  across local and remote rather than special-casing remote.
- No production read of `serverStore.remoteCaps`; no second writer; no fork of
  `resolveRemoteCaps` / `fetchRemoteModelCaps` / `deriveListCaps`.
- **The capability-age trap is not sprung** — nothing infers "this server can count" from a
  `/props`-populated record.
- `effectiveContextLength` over `contextLength` is deliberate and right: `BannerRow` already
  reads it on `main`, and `contextLength` for a local model is the GGUF-declared window, not
  the loaded `n_ctx`.
- The deferred half did not leak — no `buildChatCompletionRequest`,
  `maybeCountPromptTokens`, `occupancy` or `countUnsupported`; no request-body change, so
  the existing decisions' `usage` rationale stays intact.
- Degradation is monotone: an older build without `cache_n` yields exactly the prior value,
  never below it. `undefined` cannot render as `0` — `ratio` is computed only inside the
  `used !== undefined` block.
- The commit `79ad2e43` mutation claim ("4096 → 0") is accurate: it fails exactly the two
  tests that commit added and nothing else.
- No test reads `/slots` after a completion.
- Old persisted `{used: 0, contextFull: true}` snapshots produce no retro-banner.
- The newly reachable remote full banner has no dead CTA: `canIncrease` is false for remote,
  so only "New chat" and "Dismiss" render.
- `banner-percent` testID reuse is safe — the resolver returns one variant per render.

## Residual risk

Two early combined runs **under `--coverage`** reported failures inside `openai.test.ts`
that five later runs did not reproduce, and the names were not captured. That suite drives
real connection and idle timers, so instrumentation load is the likely cause. It can neither
be attributed to the new tests nor ruled out. **Worth one CI re-run.**

---

a-ghorbani:
### Correction to the review above — one item withdrawn

The review's closing note claimed that showing cached tokens in the turn footer was
"currently owned by nobody". **That was wrong, and I am withdrawing it.**

The evidence I gave was accurate: `cache_n` appears in this app only in `src/api/openai.ts`
and `src/utils/completionTypes.ts`, and nowhere on `main`. The conclusion did not follow.
Another open PR renders it — `AssistantTurnFooter.tsx` gates on `timings?.cache_n != null`
and renders a cached-tokens string — and it is absent from `main` for the ordinary reason
that its PR has not merged yet.

**I checked this branch and `main`, then claimed "nowhere in the app".** The gap was every
in-flight branch. On a repo with concurrent work, absence from `main` is what an unmerged
PR looks like.

Nothing in the review's findings depends on this note, and the verdict is unchanged. One
consequence is worth stating for anyone reading the architecture docs alongside this PR:
those docs are updated when code is *written*, not when it *merges*, so a statement there
may describe behaviour that is real on a branch and not yet on `main`.

---

a-ghorbani:
## Re-review after the fix round

**Verdict: REQUEST_CHANGES** — `blocker 1 (carried) · concern 0 · suggestion 4`.

**All three concerns are closed and no regression.** At the time of this re-review the
visual-evidence requirement was still outstanding, which is the only reason the verdict
was not clean. Those captures are now posted on this PR.

### The concerns, verified by mutation rather than by reading

| mutation applied to the shipped code | result |
| --- | --- |
| revert the guard to its pre-fix form | **2 tests fail** — both newly added |
| weaken the wire guard to `prompt_n` alone | **1 test fails** — exactly the new `(only cache_n)` case |
| the converse, `cache_n` alone | **3 tests fail**, incl. the existing `(only prompt_n)` |

Both arms of the guard are now mutation-killed. The first round had proved the opposite:
applying the fix left the entire suite passing, so nothing pinned either the old or the new
behaviour.

Dropping the origin qualifier from the guard was traced to the native source rather than
assumed: all three result builders set `tokens_evaluated` from the whole prompt
unconditionally, so the removed branch was unreachable locally and no local behaviour moves.
No cache term leaked into any local summation — the only summation lives in the remote API
layer, so there is no double-count.

### Gate

eslint 0 · typecheck 0 · **274 suites / 4398 passed / 2 skipped**, reproducing under
coverage · coverage well above floor · no native or Gradle path in the commit range.

### Evidence

The before/after captures are posted on this PR. They were taken against a real
`llama-server` on a physical device, with turn 1 as a control that must *not* move.

---

a-ghorbani (commented):
lgtm

PR #895, opened 2026-09-04
[Feat]: forward all samplers and parse full /props for llama.cpp

a-ghorbani:
Remote (OpenAI-compatible) servers get the sampler and reasoning controls they actually understand, one wire-name table per server type, and a settings sheet that stops displaying numbers the server is not using.

## What changed

**One profile per server type.** `src/api/servers/index.ts` holds `SERVER_PROFILES`, a `Record<ServerType, ServerProfile>` written with `satisfies`, so a server type offered in the UI cannot compile without a profile. A profile is pure data plus pure functions: `sendNames` (present = forwarded, value = wire name), `reasoningExtras`, `hasProps`, `readListRow`. `readFinish` is one module function every type shares, because field presence decides the read, not the type. No store, no clock, no network.

**Every type decision goes through `profileFor`.** The only input is the persisted `serverType`, normalised by `toServerType` at every writer. A legacy `''`, a case variant like `LLAMA.CPP`, or a free string is `unknown` and speaks the base profile. Outside `src/api/servers/` there is no literal compare or `switch` on a server-type string; a `no-restricted-syntax` pair enforces it.

**A sampler the user never moved is left to the server.** Forwarding a sampler overrides whatever the server was launched with, so `pickSamplers` omits a control whose stored value still equals the app default. `--top-k 20` on the server now survives a user who never opened the sheet.

**The settings sheet says what runs.** Three states per row: on the app default with a known server default → "Using server default: 20", no reset; changed → "Server default: 20" with reset; unknown → nothing. Reset stores the *app* default, which omits the param again, rather than pinning the server's number.

**One probed record.** `/props` resolves `contextLength`, `supportsVision` and `samplerDefaults` into `RemoteModelCaps`. An earlier round split these across three records by how long each fact stays true; every consumer read one of them, so the split cost a type and a resolver per tier and bought nothing.

## The renames

llama.cpp spells four of these differently from the app, and a wrong name is a silent no-op — the server answers 200 and ignores the field.

| App | Wire (llama.cpp) |
| --- | --- |
| `penalty_last_n` | `repeat_last_n` |
| `penalty_repeat` | `repeat_penalty` |
| `penalty_freq` | `frequency_penalty` |
| `penalty_present` | `presence_penalty` |
| `n_predict` | `max_completion_tokens` |

`top_k`, `min_p`, `typical_p`, `xtc_threshold`, `xtc_probability`, `seed`, `n_probs`, `mirostat`, `mirostat_tau`, `mirostat_eta` go under their own names. Every other server type gets the base three only (`temperature`, `top_p`, `n_predict`) until its names are verified live — vLLM spells at least one differently (`repetition_penalty`), so it is deliberately not guessed.

The `/props` read map is *derived* from the send map (`PROPS_READ_NAMES = {...sendNames, n_predict: 'n_predict'} satisfies Record<SamplerParam, string>`), so send and read cannot drift, and a new parameter that llama.cpp has no name for is a compile error.

## Which samplers are conditional

Always sent, whatever they hold: `temperature`, `top_p`, `n_predict`. These shipped unconditionally before this change, and the app's temperature default (0.7) is not the server's — OpenAI's is 1.0 — so omitting one at the app default would change what runs rather than preserve it.

Conditional (omitted at the app default), 14: `top_k`, `min_p`, `typical_p`, `xtc_threshold`, `xtc_probability`, `penalty_last_n`, `penalty_repeat`, `penalty_freq`, `penalty_present`, `mirostat`, `mirostat_tau`, `mirostat_eta`, `seed`, `n_probs`.

## Wire evidence

A wrong wire name cannot be caught by a unit test alone: llama-server returns 200 on unknown body keys, so a misspelled field ships green. The names were read back from `GET /slots?model=<id>` after a completion **the app itself built** — captured at the transport boundary and replayed, because a hand-built curl only proves the server accepts a name, not that the app sends it.

Observed for an app-assembled body: `repeat_last_n: 41`, `repeat_penalty: 1.11`, `top_k: 11`, `min_p: 0.11`, `seed: 12345`, `temperature: 0.33`, `n_probs: 2`, and `n_predict: 8` from `max_completion_tokens` — the one asymmetric pair, and so the one most likely to break unnoticed.

## Limits and deliberate trades

- **"Equals the app default" is a proxy for "the user never changed it."** Someone who deliberately sets `top_k` to exactly 40 gets no `top_k` on the wire and the server's value applies. An explicit per-parameter "use the server's value" state would fix this properly and is a separate change to the settings model.
- **`seed`** is the one conditional parameter whose omission is not observable as a value change: an untouched `-1` and an absent `seed` both mean "pick a random one".
- **The probed record is additive, so there is nothing to migrate.** `main` already persisted one `remoteCaps` record per model; this adds `samplerDefaults` to it. An entry stored by a released build hydrates with the new field absent and is filled by the next probe. (An earlier revision of this description claimed a lost-and-re-probed migration; that was wrong, and review caught it.)
- **Probes are not coalesced.** Two probes for the same server, model and url in one tick issue two requests. Callers invoke the probe detached and the write is idempotent, so the cost is the extra request.
- **The bare `/props` retry fires less often.** With one record, a scoped answer carrying only sampler defaults counts as resolved, so `contextLength` can stay unknown where the three-record version filled it in.
- **`vLLM` sampler names are unverified**, so nothing beyond the base three is sent there.
- Streaming delta shape is still read transport-side, and the sampler vocabulary is still bounded by `keyof CompletionParams`. Both are deferred by decision.

## Notes for review

- The ESLint exemption is on `src/api/servers/detect.ts` alone, not the profile table. The table uses server-type literals only as object keys, so the ban stays live inside it; `detect.ts` is the one file that genuinely compares a literal.
- The 318-row request-body snapshot matrix that guarded the restructure has been replaced by value-level tests with one owner per fact. Proving each replacement could fail turned up a guard that was never covered: `fetchServerProps`'s non-2xx early return survived deletion, because the error fixture describes no model and parses to `{}` either way. It is pinned now with a 503 carrying a model-describing body.

## Also in this change

- **The settings sheet's server-default row appears only where it applies.** It is absent from the Pal editor and from the preset sheet — those settings also govern models that have no server, and the reset there would have written one server's number into them.
- **Capabilities are re-probed once after the first completion.** On a llama.cpp router that does not autoload, the probe fired at activation asks for a model that is not resident yet, so it cannot land and every sampler row stays blank. After one completion the model is guaranteed resident. The re-probe is once per binding and does not block the turn.
- **A save that changes a server's url or type refetches its model list**, so discovery is not left empty until the next time the app is foregrounded.
- **The assistant turn footer** gains prompt speed and cached-token parts for remote turns, both origin-gated and read only when the server reported a finite number.

## Known gaps

- When a probe never landed, or the server reports a default the app's own range refuses, the row is silent: it shows the app's number with no annotation while the wire omits the parameter and the server's value runs.
- `removeUserSelectedModel` does not prune that model's capability entry (pre-existing; the stored payload is now slightly larger).
- If the post-completion re-probe itself fails, recovery is again only on foreground. Retrying per turn would mean a probe every turn against a server that is down.


a-ghorbani:
Posted as a comment rather than a formal **Request changes** review: the bot identity was unavailable (`GH_APP_ID` unset), so this ran under the PR author's own account and GitHub refuses a self-review. The verdict below is REQUEST_CHANGES and should be read as one.

# Independent review

verdict: **REQUEST_CHANGES**
role_subreviews: **COMPLETED** (8/8 + an independent refutation pass)
review_complete: **yes**
scale: 29 files, +2932 / −540

No blocker in the code. One blocking process condition, seven concerns, and a
long tail of suggestions. The change is well built and unusually well evidenced;
most of what follows is about reach beyond the stated scope, not about the wire
work itself.

---

## Blocking

**BL1 — No visual evidence on the PR.** `gh pr view 895` → no images in the body,
zero comments. `AGENTS.md` makes this blocking on its own for a change with
visible UI.

The cause is environmental, not authorial: `tools/post-pr-visual-evidence.sh`
exits 3 with `MANUAL_POST_REQUIRED` — no GitHub session token for the image
upload. Four captures exist and were verified by reading them.

It is worse than a missing upload, because of C2 below: the four captures cover
the **remote** surfaces only, and this change also alters the **local** turn
footer, for which there is no capture at all and no mention in the PR body.

---

## Concerns

**C1 — "Reset to server default" writes a value the app's own validator then
rejects.** `src/components/CompletionSettings/CompletionSettings.tsx:83-95`.

Found independently by five reviewers. The reported default is passed to
`onChange` unclamped, while the same component holds the metadata that bounds
the control. `--top-k 0` (the standard way to disable top-k) and
`--repeat-last-n -1` are ordinary llama.cpp configurations; the app's ranges are
`top_k` min 1 and `penalty_last_n` min 0. Proven by probe:

```
onChange: ('top_k', 0), ('penalty_last_n', -1)
validateCompletionSettings → isValid:false
  top_k: "Value must be between 1 and 128"
  penalty_last_n: "Value must be between 0 and 256"
```

The chip renders and presses; Save then aborts with an alert naming a parameter
the user never typed, with no highlight on the offending control and no clean
recovery but discarding the draft. Once written, the tolerance check matches, so
the affordance collapses to the plain "server default" label while the clamped
slider shows a different number.

Fix: return `null` when the reported value fails the control's own validation.

**C2 — The "remote turn footer" change fires on every local turn, undescribed,
untested and uncaptured.** `src/components/AssistantTurnFooter/AssistantTurnFooter.tsx:66-81`.

Raised by two reviewers; I verified it directly and both understated it. The
footer gates on **presence** (`!= null`), not origin. llama.rn declares
`NativeCompletionResultTimings` with **both** `cache_n` and `prompt_per_second`
as required, non-optional numbers, and `useChatSession.ts` spreads engine
timings with no origin branch. Neither field renders on `origin/main`.

So every **local** turn footer gains **two** new metrics, not one. Both new
footer tests build from the remote finish-chunk fixture, and every local-turn
fixture omits the fields, so nothing asserts the local string.

Fix: decide deliberately — either gate on origin, or accept the local change,
describe it in the PR body, test it, and capture it.

**C3 — Server sampler defaults are rendered on, and reset into, two scopes that
are not the session's model.** `PalGenerationSettingsSheet.tsx:252`,
`ChatGenerationSettingsSheet.tsx:362`.

`ModelStore.activeSamplerDefaults` resolves against `activeModel` and the live
binding. Its binding guard is careful on the axis it checks, but both call sites
hand the result to a surface editing a different scope: the Pal sheet edits a Pal
carrying its own `defaultModel`, and the chat sheet edits the app-wide preset
whenever `isEditingPresetSettings`. One server's numbers are therefore labelled
"server default" for, and reset into, settings belonging to a different model —
and the row appears or vanishes with whatever happens to be loaded.

`context/architecture/remote-servers.md:1163` asserts the opposite for the Pal
case; the preset case has no row at all.

**C4 — The volatile tier is structurally assignable into both persisted tiers.**
`src/utils/types.ts:536-540`.

`RemoteModelPresence` carries no `tier` field, and every field of both persisted
types is optional. Verified under `tsc --noEmit --strict`: props→caps, caps→props
and the corresponding `Record` conversions all error; **presence→caps,
presence→props and `Record<string,Presence>`→`Record<string,Caps>` produce no
diagnostic**. `this.remoteCaps[key] = presence` compiles and would persist a
volatile claim.

`remote-servers.md:419-423` (restated at `:886`) publishes the opposite as a
structural invariant. Fix: add `tier?: 'presence'`, and correct the doc.

**C5 — Unbounded wire strings are persisted, and on Android one probe can break
persistence of the whole store.** `src/api/openai.ts:414`, `:550-558`;
`ServerStore.ts:127-137`.

`nonEmptyString` applies no length cap to `build_info` or `model_alias`. Every
sibling field is properly bounded. Survived an adversarial attack and came back
stronger: driving the real `fetchServerProps` with a 4 MB `build_info` produced a
persisted blob of **8,388,648 bytes** — one probe of one model, no accumulation.
The Android ceiling is real and traced through the vendored source
(`config.gradle` `dbSizeInMB = 6L` → `ReactDatabaseSupplier.setMaximumSize`, no
override in `android/gradle.properties`). Because `makePersistable` writes one
record for the whole store, `servers` and `privacyNoticeAcknowledged` go down
with it.

Two qualifications from the refutation, both material: the consequence is
**Android-only** (iOS spills values over 1 KB to uncapped files), and unbounded
wire strings already reach the blob on main via `RemoteModelInfo.id`. But
`remoteProps` is new here, and nothing reads either field today — so capping or
dropping them is nearly free.

**C6 — The new probe coalescing is a PR-introduced regression on two recovery
paths.** `ServerStore.ts:82-84, 373-391`.

`probesInFlight` is absent from main. Proven by test: a second caller arriving
while a probe hangs issues zero new HTTP requests and writes nothing. Two
consequences:

- The iOS foreground reprobe (`ModelStore.ts:1214-1215`) exists precisely because
  the first probe is the request that raises the local-network prompt, so a grant
  always arrives after it has already failed. The reprobe now joins the doomed
  request instead of issuing one under the new grant. The window is ~10 s, not 5
  — `probeRemoteModel` makes two sequential `fetchServerProps` legs and the
  `.finally` hangs off the whole thing.
- The dedup key carries no url, so a refresh issued after a server url edit joins
  the pre-edit promise: `fetchServerProps calls: 1`, `urls asked:
  ["http://old:8080"]`, both maps empty, and the caller resolves as success.

Impact is narrower than first reported: the reprobe fires on *every* foreground
transition and a settled probe releases the entry, so it self-heals on the next
cycle rather than staying unknown for the session. Fix: clear the map on leaving
`active`, or add a `force` flag; and put the url in the key.

**C7 — `PROPS_SCALAR_FIELDS` reintroduces exactly the drift `CAPS_FIELDS` exists
to prevent.** `ServerStore.ts:41-49`.

Two hand-maintained lists sit directly under a comment reading *"Enumerated once
so the usability check and the no-op write check cannot drift apart."* Adding a
scalar to `PROPS_FIELDS` alone makes `samePropsContent` report `true` for a
changed answer, and the write is silently skipped — in a map this PR has just
made persistent.

---

## Suggestions

| # | Finding | Where |
| --- | --- | --- |
| S1 | Footer `.toFixed()` on never-validated wire numbers throws in render with no error boundary, and the value persists so it recurs on reopen. **Downgraded**: a hostile server already has the identical crash+persist primitive on main via `predicted_per_second`; this adds a third key to a saturated primitive. Still one line to fix with the existing `finiteNumber` helper. | `AssistantTurnFooter.tsx:66-81`, `openai.ts:1009` |
| S2 | The reset writes the raw double while the label rounds, so `0.800000011920929` resurfaces in the numeric box on the next sheet open. The PR's own test pins the current behaviour. | `CompletionSettings.tsx:85` |
| S3 | The reset affordance: no `accessibilityRole="button"` (159 precedents in-repo), a ~16-20 dp target that silently spans the full card width with no `hitSlop`, `theme.colors.primary` is body-grey in this app so nothing looks tappable, and `disabled` renders identically. | `CompletionSettings.tsx:84-93`, `styles.ts:27-30,73-77` |
| S4 | The per-`serverType` payload table — the one doc that governs "is this field safe for this server" — was not updated for `reasoning_budget_tokens`. | `openai.ts:669-687` |
| S5 | The whole presence tier and four of five `remoteProps` fields have no non-test consumer, and four of them are now persisted. Persisted schema is the expensive kind to carry. | `types.ts:516-540` |
| S6 | `removeUserSelectedModel` prunes none of the four per-model maps; a removed-then-re-added model resurrects its old entry. | `ServerStore.ts:208-212` |
| S7 | `FORWARD_ALLOWLIST` admits `n_predict`, whose send name is `max_completion_tokens`; inert only because `StreamChatParams` has no such field. `type ForwardableParam = Exclude<SamplerParam,'n_predict'>`. | `openai.ts:112` |
| S8 | `utils/types.ts` importing `SamplerParam` from `api/openai`. **Downgraded**: type-only, all fourteen imports erase, nine pre-existing precedents, and the proposed move splits llama-server HTTP vocabulary into a llama.rn module and separates it from `FORWARD_ALLOWLIST`. | `types.ts:14` |
| S9 | `activeSamplerDefaults` depends on the whole `capabilityEnv` while reading two of seven fields, dragging an O(servers × models) rebuild along. Zero incremental cost today because the chain is already observed. | `ModelStore.ts:2607-2610` |
| S10 | Presence is the one tier written with no no-op guard, and it is minted with a fresh `at` on every probe. Free today; the asymmetry is invisible to the next author. | `ServerStore.ts:481-483` |
| S11 | Fixture provenance is unenforceable from this repo — the captures live in the dev-team repo, so no CI check can catch a future hand-edit. The header's "not reformatted" is also literally untrue (JSON→TS transcription), though the substance verified. A local HF cache path is committed in `model_path`. | `jest/fixtures/llamaServerWire.ts` |
| S12 | Two comments justify declarations on "nothing does this yet" grounds — case 3 under the four-case test, and symptomatic of S5. | `listCaps.ts:16`, `modelCaps.ts:26` |
| S13 | `renderServerDefault(name, 0)` at `CompletionSettings.tsx:152` is unreachable. `FORWARD_ALLOWLIST` and `REASONING_BUDGET_TOKENS` are keyed by `string` where `ServerTypeOption` and `EffortLevel` exist — an added effort level fails silently to `-1`, *uncapped*. | various |
| S14 | Localization polish: `"{{value}} cached"` drops its noun; casing flips between the two states of the same row. | `en.json:567,688-689` |

---

## Verified clean — recorded so a follow-up pass does not re-derive it

- **Fixture provenance.** Two reviewers independently deep-compared all five
  exports against the capture files: identical, key order included, down to the
  6 KB Jinja template and unused `lora`/`speculative` noise. Projection is real
  (`penalty_last_n: wire.repeat_last_n`,
  `Object.keys(slotsAfterSamplerRequest[0].params)`), not decorative.
- **`PARAM_WIRE_NAME` 16/16**, checked against slot `id: 3` of the read-back
  capture — every name confirmed *accepted*, not merely present. The four
  `penalty_*` spellings appear in neither capture, which positively confirms the
  renames are required.
- **`FORWARD_ALLOWLIST` keys** match `SERVER_TYPE_OPTIONS` and
  `detectServerType` exactly.
- **Falsifiability.** Six mutations across two reviewers, every one caught by a
  named test, all files restored.
- **Hygiene.** Zero internal tracker IDs or story anchors across every added
  line and all 17 commit messages.
- **Comments.** 160 added lines against 2513 added non-whitespace (6.4%),
  overwhelmingly case 4 — proportionate for a wire-contract boundary. Three
  exceptions flagged by case number.
- **Hydration and downgrade.** An absent `supportsAudio` reads `'unknown'`
  end-to-end, never a definite `false`; pre-`probedUrl` entries are dropped
  whole; rollback is safe with no migration.
- **Key shape** has no collision and no cross-server prefix match.
- **`remotePresence` is genuinely absent** from the `makePersistable` list,
  verified in the call rather than the comment.
- **Nothing added to the streaming hot path**; no bundle-size change; no new
  endpoint, permission or manifest surface; the only wire value reaching a URL is
  `encodeURIComponent`-encoded.
- **`NATIVE_CHANGES=NO`**, re-derived from the diff by two reviewers with the
  commit-scoped check. `pod install` and the two platform builds are not
  required.

## Verification

`tsc` clean · `eslint` 0 errors on PR files · `jest` 275 suites / 4451 passed /
2 skipped · `l10n:validate` valid · native not required. Full detail, including
a lint false positive caused by a concurrent reviewer's scratch file, is in
`verification.md`.

## Residual risk

- Behaviour on a llama-server build newer than `b9976-e3546c794` is unverified.
  The author states this limit rather than generalising from it, and the failure
  mode for every field except the sampler names is fail-safe.
- vLLM's acceptance of the sampler set is unverified; its allow-list row is
  empty, so nothing is sent.
- No device run beyond the four captures, none of which covers the local footer
  change in C2, or Android/iOS divergence on the new row.

## Open questions for the author

1. Is the local-turn footer change (C2) intended? The PR presents the footer work
   as remote-only.
2. Should the presence tier and the four unread `remoteProps` fields ship now
   (S5), given they are persisted and have no consumer in this change?
3. C3 asks a product question, not just a code one: what should a Pal or preset
   sheet show when the "server default" belongs to a model that is not the one
   being configured?

---

Eight role reviewers plus an adversarial refutation pass, none of them the implementer, against a separate worktree built from the PR ref. Two single-sourced concerns were downgraded by the refutation and one had its impact corrected; the surviving ones are recorded with what was done to try to break them.

a-ghorbani:
## Two review findings fixed

**The footer parts are now gated on origin, not presence** (`bc8ec498`).

llama.rn declares both `cache_n` and `prompt_per_second` as required,
non-optional numbers on local completions, and neither renders on `main` — so
gating on presence silently added **two** metrics to every local turn footer.
That was an accident, not a decision: undescribed, unasserted and uncaptured.
The local footer is worth having, but it deserves its own change.

The test asserts both arms from one `timings` object — suppressed on a local
turn, present on a remote one. A screenshot could not carry this: the claim is
an absence, and an absence in a capture cannot be told apart from a stale build
or a turn that never populated timings.

**"Reset to server default" no longer offers a value saving would reject**
(`76ce3ec7`).

`--top-k 0` is the ordinary way to disable top-k and `--repeat-last-n -1` is an
ordinary setting; this app's ranges start at 1 and 0. The chip rendered, pressed,
and then Save aborted with an alert naming a parameter the user never typed, with
no route back but discarding the draft. An out-of-range default is now withheld
entirely — claiming it as "server default" would be equally wrong.

Both fixes were checked by reverting them: exactly two tests fail, one per fix,
and pass again when restored.

Gates: 274 suites, 4448 passed, 2 skipped. Typecheck and lint clean.

The remaining review findings are unaddressed and still stand — in particular the
server defaults presented on Pal and preset sheets that configure a different
model, and the unbounded persisted `build_info` / `model_alias` strings.

Visual evidence is still not posted; that is an environment blocker, not an
authoring one.

